PR #21, opened 2021-08-06
feat: loginWithRedirect and handleCallback

wangsijie:
Tests are desigend for debugging(run through) now, will be changed to unit test and e2e test in the future.


packages/client/__tests__/index.spec.ts
```diff
@@ -44,4 +45,20 @@ describe('init client', () => {
     const configuration = client.issuer.metadata;
     expect(configuration.authorization_endpoint).toContain('oidc/auth');
   });
+  test('get login url and codeVerifier', () => {
+    const [url, codeVerifier] = client.getLoginUrlAndCodeVerifier();
+    console.log(url);
+    console.log(codeVerifier);
```
simeng-li, line 51: nit: do we need to keep these logs

simeng-li (approved):
LGTM

PR #30, opened 2021-09-27
feat: redirect and login

wangsijie:
[LOG-125](https://linear.app/silverhand/issue/LOG-125/redirect-and-loginget-token)


packages/client/src/jwt.ts
```diff
@@ -24,8 +24,13 @@ export const fetchJwks = async (url: string): Promise<JWK[]> => {
 };
 
 export const decode = (token: string) => {
-  const { payload } = jwt.decode(token, { complete: true });
-  const { iss, sub, aud, exp, iat, at_hash } = payload;
+  const jwtDecoded = jwt.decode(token, { complete: true });
```
simeng-li, line 27: just wondering if jwt.decode throws any exceptions that need to be caught ? By default it returns falsy/empty value?

packages/client/src/index.ts
```diff
@@ -1,12 +1,11 @@
...
 
 export interface ConfigParameters {
-  discoveryUrl: string;
+  logtoUrl: string;
   clientId: string;
-  redirectUris?: string[];
 }
 
-export const extractBearerToken = (authorization: string): string => {
+export const extractBearerToken = (authorization: string): string | null => {
```
simeng-li, line 8: will throw empty string work?

packages/client/src/index.ts
```diff
@@ -20,53 +19,52 @@ export const extractBearerToken = (authorization: string): string => {
...
-
   return {
     clientId,
-    discoveryUrl,
-    redirectUris,
+    logtoUrl,
   };
 };
 
+export const ensureTrailingSlash = (url: string): string => {
```
simeng-li, line 37: should we keep a util or lib files for these methods?

packages/client/src/index.ts
```diff
@@ -89,17 +83,14 @@ export class LogtoClient {
       prompt: 'consent',
       code_challenge: codeChallenge,
       code_challenge_method: 'S256',
+      redirect_uri: redirectUrl,
     });
     return [url, codeVerifier];
   }
 
-  public async handleLoginCallback(codeVerifier: string, code: string) {
+  public async handleLoginCallback(redirectUrl: string, codeVerifier: string, code: string) {
```
simeng-li, line 91: redirectUri?

packages/client/src/index.test.ts
```diff
@@ -34,19 +34,18 @@ describe('init client', () => {
...
   test('openid configuration', async () => {
-    const configuration = client.issuer.metadata;
-    expect(configuration.authorization_endpoint).toContain('oidc/auth');
+    const configuration = client.issuer?.metadata;
+    expect(configuration?.authorization_endpoint).toContain('oidc/auth');
   });
   test('get login url and codeVerifier', () => {
-    const [url, codeVerifier] = client.getLoginUrlAndCodeVerifier();
+    const [url, codeVerifier] = client.getLoginUrlAndCodeVerifier('http://localhost:3000/callback');
     console.log(url);
```
simeng-li, line 49: should we remove the console.log statement?

packages/client/src/index.test.ts
```diff
@@ -58,7 +57,11 @@ describe('init client', () => {
       expect(1).toEqual(1);
```
simeng-li, line 57: ?

packages/client/src/index.test.ts
```diff
@@ -34,19 +34,18 @@ describe('init client', () => {
...
-    expect(configuration.authorization_endpoint).toContain('oidc/auth');
+    const configuration = client.issuer?.metadata;
+    expect(configuration?.authorization_endpoint).toContain('oidc/auth');
   });
   test('get login url and codeVerifier', () => {
-    const [url, codeVerifier] = client.getLoginUrlAndCodeVerifier();
+    const [url, codeVerifier] = client.getLoginUrlAndCodeVerifier('http://localhost:3000/callback');
     console.log(url);
     console.log(codeVerifier);
     expect(typeof url).toEqual('string');
```
simeng-li, line 51: since we have defined the client.getLoginUrlAndCodeVerifier's return type as [string, string]. Feels like no need to validate within ut case here?

simeng-li (commented):
leave some comments please check thx

packages/client/src/index.ts
wangsijie: I raised this question in `logto`, can't find the link for now. The conclusion is to do it later.

packages/client/src/jwt.ts
wangsijie: No, it won't throw, just return null.

packages/client/src/index.test.ts
wangsijie: It is used for manual test, automatic browser test is not implemented for now.

packages/client/src/index.test.ts
wangsijie: same problem to `console.log`, the tests here need to be run twice, is dummy, will fix it someday

packages/client/src/index.test.ts
wangsijie: yep, useless validation here, just for the `console.log`

wangsijie:
the unit tests on loginRedirect here are stupid, but it's hard to do it, let's leave it here for now.

wangsijie:
in addition, a headless chrome will be needed to implement the auto test. I think it can be done someday later.

simeng-li:
thx. overall LGTM.  defer the final stamp to @gao-sun as might have more context & guidelines at the current  initial stage.

packages/client/src/index.test.ts
gao-sun: looks like it's not about manual/auto test? should expect the url starts with the same host and `codeVerifier` has the correct format

packages/client/src/index.test.ts
```diff
@@ -10,22 +10,18 @@ describe('extractBearerToken', () => {
...
+    expect(() => extractBearerToken('testtoken')).toThrow();
   });
   test('empty string', () => {
-    const token = extractBearerToken('');
-    expect(token).toBeNull();
+    expect(() => extractBearerToken('')).toThrow();
   });
   test('empty input', () => {
     // eslint-disable-next-line @typescript-eslint/prefer-ts-expect-error
     // @ts-ignore
```
gao-sun, line 23: should use `// @ts-expect-error`

packages/client/src/index.test.ts
gao-sun: looks like these `console.log` doesn't make sense in UTs. we need always expect something.

packages/client/src/index.test.ts
```diff
@@ -58,7 +53,11 @@ describe('init client', () => {
       expect(1).toEqual(1);
     }
 
-    const tokenset = await client.handleLoginCallback(process.env.CODE_VERIFIER, process.env.CODE);
+    const tokenset = await client.handleLoginCallback(
+      'http://localhost:3000/callback',
+      process.env.CODE_VERIFIER || '',
+      process.env.CODE || ''
+    );
```
gao-sun, line 60: why do we need envs when we have `codeVerifier` generated? UTs shouldn't involve envs IMO.

packages/client/src/index.ts
```diff
@@ -12,61 +11,60 @@ export const extractBearerToken = (authorization: string): string => {
     typeof authorization !== 'string' ||
```
gao-sun, line 11: redundant type check?

packages/client/src/index.ts
```diff
@@ -12,61 +11,60 @@ export const extractBearerToken = (authorization: string): string => {
...
   return token;
 };
 
 export const ensureBasicOptions = (options?: ConfigParameters): ConfigParameters => {
-  const { clientId, discoveryUrl, redirectUris } = options || {};
-  if (typeof discoveryUrl !== 'string' || !discoveryUrl.startsWith('http')) {
-    throw new Error('Invalid discoveryUrl');
+  const { clientId, logtoUrl } = options || {};
+  if (typeof logtoUrl !== 'string' || !logtoUrl.startsWith('http')) {
+    throw new Error('Invalid logtoUrl');
```
gao-sun, line 24: 1. "ensure" sounds weird in general. maybe "validateSomething"
2. can we align the usage of "options" "config" and "parameters"? the naming is confusing now. `Options` is good to go
3. went through the code i think there's no need to validate the config since we're writing TypeScript. we should always treat the input valid unless needed (like post body in APIs). if we have to validate something, we can just leverage `zod` like logto-io/core.

packages/client/src/index.ts
gao-sun: maybe "appendSlashIfNeeded" ? 

P.S. we can somehow follow [Swift API Design Guideline](https://swift.org/documentation/api-design-guidelines/) cc @simeng-li @xiaoyijun

gao-sun: created [LOG-130](https://linear.app/silverhand/issue/LOG-130/standardize-dev-pipeline-for-silverhandessentials) for this

packages/client/src/index.ts
```diff
@@ -12,61 +11,60 @@ export const extractBearerToken = (authorization: string): string => {
...
-    void this.initIssuer(discoveryUrl, onOidcReady);
+    void this.initIssuer(
+      `${ensureTrailingSlash(logtoUrl)}oidc/.well-known/openid-configuration`,
+      onOidcReady
+    );
   }
 
   public getClient(): Client {
+    if (!this.issuer) {
+      throw new Error('should init first');
```
gao-sun, line 62: shall we use codes (constants) for errors first (like extract the constants [in a file](https://github.com/logto-io/logto/blob/master/packages/phrases/src/locales/en.ts)), for easier i18n in the future?

no hurry for i18n. we can add an issue first.

packages/client/src/index.ts
```diff
@@ -89,17 +83,14 @@ export class LogtoClient {
...
 
-  public async handleLoginCallback(codeVerifier: string, code: string) {
+  public async handleLoginCallback(redirectUri: string, codeVerifier: string, code: string) {
     const client = this.getClient();
-    const tokenset = await client.callback(
-      this.redirectUris[0],
-      { code },
-      { code_verifier: codeVerifier }
-    );
+    const tokenset = await client.callback(redirectUri, { code }, { code_verifier: codeVerifier });
```
gao-sun, line 93: should be `tokenSet` per [doc](https://github.com/panva/node-openid-client/blob/main/docs/README.md#clientcallbackredirecturi-parameters-checks-extras):

> Returns: `Promise<TokenSet>` Parsed token endpoint response as a TokenSet.

gao-sun (changes requested):

gao-sun:
plus, please follow the same commitlint format for the PR title. ;-)

packages/client/src/index.test.ts
wangsijie: this redirect needs a browser (login to logto.dev), I do this by 2 steps (1. get redirect url and verifier 2. manually copy & paste to browser and get a login code, then set them to envs and test `handleLoginCallback` will change to `cypress` or sth soon.

packages/client/src/index.ts
wangsijie: some developers may use our packages without typescripts, then will the type definition still work?

packages/client/src/index.ts
wangsijie: sure, I'll create a seperate issue in linear.

packages/client/src/index.ts
gao-sun: if they are using JS, IDEs will still read type definitions for code intelligence and we'll also offer docs to instruct the developer. if they persist to ignore all these hints, they we can just let they go.

APIs should be guarded since they are public and we cannot trust any input. unlike SDKs, which are for "internal" use normally.

packages/client/src/index.test.ts
gao-sun: as discussed offline, will remove e2e part in UTs

packages/client/src/index.ts
wangsijie: agreed, type check will be removed, even if we need that in the future, an additional layer is better.

PR #33, opened 2021-10-12
refactor: use jose for jwt

wangsijie:
[LOG-155](https://linear.app/silverhand/issue/LOG-155/jwt%E7%BB%9F%E4%B8%80%E7%94%A8jose%E5%A4%84%E7%90%86)


gao-sun (approved):
lgtm

packages/client/src/index.test.ts
```diff
@@ -58,29 +59,26 @@ const generateRandomString = (length: number) =>
...
   beforeAll((done) => {
     client = new LogtoClient(
       {
         logtoUrl: 'https://logto.dev',
         clientId: 'foo',
       },
-      () => {
+      async () => {
+        id_token = await new SignJWT({})
+          .setProtectedHeader({ alg: 'HS256' })
```
simeng-li, line 76: should we use RS256 sign algorithm by default?

wangsijie: fixed, although it doesn't matter in this case, but rs256 would be better.

PR #34, opened 2021-10-13
feat: decode token

wangsijie:
I am doing a large refractor to replace `openid-connect`, since it does not support browser.

This PR is a small part.


packages/client/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+  if (!payloadPart) {
+    throw new Error('invalid token');
+  }
+
+  const payloadString = payloadPart.replace(/-/g, '+').replace(/_/g, '/');
+  const json = decodeURIComponent(
+    escape(Buffer.from(fullfillBase64(payloadString), 'base64').toString())
+  );
+
+  const payload = JSON.parse(json) as IDToken;
```
gao-sun: should use `zod` to guard the json?

packages/client/src/utils.ts
```diff
@@ -0,0 +1,38 @@
...
+
+  return input;
+};
+
+/**
+ * Decode IDToken from jwt, without verifing.
+ * Verifing jwt requires fetching public key first, this can not
+ * be done in a sync function, in some cases, verifing is not necessary.
+ *
+ * @param token jwt string
```
gao-sun, line 20: let's use `JWT` for cross-repo naming convention

gao-sun: also let's follow the same format for docs
image

packages/client/package.json
```diff
@@ -13,7 +13,8 @@
   "dependencies": {
     "axios": "^0.21.1",
     "jose": "^3.14.3",
-    "openid-client": "^4.7.4"
+    "openid-client": "^4.7.4",
```
gao-sun, line 16: nit: this dependency can be removed? as well as `axios`

packages/client/src/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+
+  const payloadString = payloadPart.replace(/-/g, '+').replace(/_/g, '/');
+  const json = decodeURIComponent(
+    escape(Buffer.from(fullfillBase64(payloadString), 'base64').toString())
+  );
+
+  try {
+    const payload = JSON.parse(json) as IDToken;
+
+    return IDTokenSchema.parse(payload);
```
gao-sun, line 48: no need for `as` as zod will parse it with typed result

packages/client/src/utils.ts
```diff
@@ -0,0 +1,56 @@
...
+ * be done in a sync function, in some cases, verifing is not necessary.
+ * @param token JWT string
+ * @returns IDToken combined with JWT Claims.
+ */
+export const decodeToken = (token: string): IDToken => {
+  const payloadPart = token.split('.')[1];
+
+  if (!payloadPart) {
+    throw new Error('invalid token');
+  }
```
gao-sun, line 38: maybe no need for split payload/signing content, we can just decode and ignore sigining?

gao-sun (changes requested):

packages/client/package.json
gao-sun: ignore me about `axios`. i'm thinking about changing `core` to use `axios` for cross-repo consistency.

packages/client/src/utils.ts
```diff
@@ -0,0 +1,54 @@
...
+  at_hash: z.optional(z.string()),
+});
+
+export type IDToken = z.infer<typeof IDTokenSchema>;
+
+/**
+ * Decode IDToken from JWT, without verifing.
+ * Verifing JWT requires fetching public key first, this can not
+ * be done in a sync function, in some cases, verifing is not necessary.
+ * @param token JWT string
```
gao-sun, line 30:
```suggestion
 * @param token JWT string.
```

gao-sun (approved):

PR #35, opened 2021-10-14
feat: codeVerifier and codeChallenge

wangsijie:


packages/client/src/generators.test.ts
```diff
@@ -0,0 +1,12 @@
+import { generateCodeVerifier, generateCodeChallenge } from './generators';
+
+test('generate codeVerifier', () => {
+  const verifier = generateCodeVerifier();
+  expect(verifier.length).toEqual(96);
```
simeng-li, line 5: nit: it would be better to give all the constant number a readable const definition. 
i.e. 
```js
const CODE_VERIFIER_LEN = 96;
const HASH_ENCODED_LEN = 43
```

packages/client/src/generators.ts
```diff
@@ -0,0 +1,37 @@
+import { fromByteArray } from 'base64-js';
+import { sha256 } from 'js-sha256';
+
+function generateRandomData(length: number) {
```
simeng-li, line 4: we could have an open issue on this one to discuss if there is a better or more elegant approach.

simeng-li (approved):
overall LGTM. let's discuss offline on this and refactor later if needed.

packages/client/src/generators.ts
```diff
@@ -0,0 +1,37 @@
...
+  // Use web crypto APIs if possible
```
gao-sun, line 5: how about we just use crypto directly like [playground](https://github.com/logto-io/playground/blob/master/src/pages/Home/index.tsx#L28), and do backward compatibility work afterwards? actually the web crypto api is [widely supported](https://developer.mozilla.org/en-US/docs/Web/API/Web_Crypto_API#browser_compatibility) (even in IE 11)

gao-sun: and no need for `js-sha256` dependency since using subtle is enough

gao-sun (changes requested):

PR #36, opened 2021-10-14
chore: remove extractBearerToken

wangsijie:
`@logto/client` is focused on browser, remove `extractBearerToken`

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -6,18 +6,6 @@ export interface ConfigParameters {
   clientId: string;
 }
 
-export const extractBearerToken = (authorization: string): string => {
-  if (
-    !authorization ||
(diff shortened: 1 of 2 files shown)


packages/client/src/index.ts
```diff
@@ -6,18 +6,6 @@ export interface ConfigParameters {
   clientId: string;
 }
 
-export const extractBearerToken = (authorization: string): string => {
-  if (
-    !authorization ||
-    (!authorization.startsWith('Bearer ') && !authorization.startsWith('bearer '))
```
simeng-li, line 12: nit: Just  FYI If I remember correctly `authorization` header namespace as well as `Bearer` keyword are case insensitive.  In other words, feels like `BEARER` might be valid as well.  Better use `toLowerCase()` here.

PR #37, opened 2021-10-14
feat: discover

wangsijie:
TBD: network error handling: [LOG-161](https://linear.app/silverhand/issue/LOG-161/op-request-%E9%94%99%E8%AF%AF%E5%A4%84%E7%90%86)

### DIFF
--- a/packages/client/src/discover.ts
+++ b/packages/client/src/discover.ts
@@ -1,0 +1,27 @@
+import axios from 'axios';
+import { z } from 'zod';
+
+const OIDCConfigurationSchema = z.object({
+  authorization_endpoint: z.string(),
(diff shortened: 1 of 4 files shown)


packages/client/src/discover.ts
```diff
@@ -0,0 +1,27 @@
...
+const appendSlashIfNeeded = (url: string): string => {
+  if (url.endsWith('/')) {
+    return url;
+  }
+
+  return url + '/';
+};
+
+export default async function discover(url: string): Promise<OIDCConfiguration> {
+  const { data } = await axios.get<OIDCConfiguration>(
```
simeng-li, line 23: nit: could have inline `TODO:` comments here as well.

PR #39, opened 2021-10-14
feat: verifyIdToken

wangsijie:
## Summary

verify an id_token JWT

### DIFF
--- a/packages/client/jest.setup.js
+++ b/packages/client/jest.setup.js
@@ -1,6 +1,8 @@
+// Need to disable following rulus to mock text-decode/text-encoder and crypto for jsdom
+// https://github.com/jsdom/jsdom/issues/1612
 /* eslint-disable node/prefer-global/text-decoder */
-/* eslint-disable @silverhand/fp/no-mutation */
 /* eslint-disable node/prefer-global/text-encoder */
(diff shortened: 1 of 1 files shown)


packages/client/src/verify-id-token.ts
```diff
@@ -0,0 +1,22 @@
...
+export const createJWKS = (JWKSUri: string) => {
+  return createRemoteJWKSet(new URL(JWKSUri));
+};
+
+export const verifyIdToken = async (
+  JWKS: GetKeyFunction<JWSHeaderParameters, FlattenedJWSInput>,
+  idToken: string,
+  audience: string
+): Promise<void> => {
+  await jwtVerify(idToken, JWKS, {
```
simeng-li, line 17: should we return the `jwtVerify` promise out?  With current approach we won't get access to the  JWTVerify return value from the upper scope? Not sure whether this is expected.

packages/client/jest.setup.js
```diff
@@ -1,9 +1,14 @@
+/* eslint-disable node/prefer-global/text-decoder */
```
gao-sun, line 1: maybe some comments for why we disable the rules?

PR #40, opened 2021-10-14
feat: grantTokenByAuthorizationCode

wangsijie:
## Summary

after PKCE redirecting back, the returning url will contain and authorization code, grant token by this code.

### DIFF
--- a/packages/client/src/grant-token.ts
+++ b/packages/client/src/grant-token.ts
@@ -1,0 +1,32 @@
+import axios from 'axios';
+import { z } from 'zod';
+
+const TokenSetParametersSchema = z.object({
+  access_token: z.string(),
+  expires_in: z.number(),
+  id_token: z.string(),
+  refresh_token: z.string(),
+});
+
+export type TokenSetParameters = z.infer<typeof TokenSetParametersSchema>;
+
+export const grantTokenByAuthorizationCode = async (
+  endpoint: string,
+  code: string,
+  redirectUri: string,
+  codeVerifier: string
+): Promise<TokenSetParameters> => {
+  const parameters = new URLSearchParams();
+  parameters.append('grant_type', 'authorization_code');
+  parameters.append('code', code);
+  parameters.append('redirect_uri', redirectUri);
+  parameters.append('code_verifier', codeVerifier);
+
+  const { data } = await axios.post<TokenSetParameters>(endpoint, parameters, {
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+  });
+
+  return TokenSetParametersSchema.parse(data);
+};
--- a/packages/client/src/grant-token.test.ts
+++ b/packages/client/src/grant-token.test.ts
@@ -1,0 +1,48 @@
+import nock from 'nock';
+
+import { grantTokenByAuthorizationCode } from './grant-token';
+
+describe('grantTokenByAuthorizationCode', () => {
+  test('get tokenSet paramaters', async () => {
+    const successResponse = {
+      access_token: 'access_token',
+      expires_in: 3600,
+      id_token: 'id_token',
+      refresh_token: 'refresh_token',
+    };
+
+    nock('https://logto.dev', { allowUnmocked: true })
+      .post('/oidc/token')
+      .reply(200, successResponse);
+
+    const tokenSet = await grantTokenByAuthorizationCode(
+      'https://logto.dev/oidc/token',
+      'code',
+      'http://localhost:3000/callback',
+      'verifier'
+    );
+
+    expect(tokenSet).toMatchObject(successResponse);
+  });
+
+  test('lack of access_token', async () => {
+    const successResponse = {
+      expires_in: 3600,
+      id_token: 'id_token',
+      refresh_token: 'refresh_token',
+    };
+
+    nock('https://logto.dev', { allowUnmocked: true })
+      .post('/oidc/token')
+      .reply(200, successResponse);
+
+    await expect(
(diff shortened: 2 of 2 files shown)


packages/client/src/grant-token.ts
```diff
@@ -0,0 +1,32 @@
...
+  parameters.append('redirect_uri', redirectUri);
+  parameters.append('code_verifier', codeVerifier);
+
+  const { data } = await axios.post<TokenSetParameters>(endpoint, parameters, {
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+  });
+
+  return TokenSetParametersSchema.parse(data);
```
gao-sun, line 31: in general, we should assume api returns the correct schema, otherwise the generic param in `axios.post<>` is meaningless.

packages/client/src/grant-token.ts
```diff
@@ -0,0 +1,32 @@
...
+  redirectUri: string,
+  codeVerifier: string
+): Promise<TokenSetParameters> => {
+  const parameters = new URLSearchParams();
+  parameters.append('grant_type', 'authorization_code');
+  parameters.append('code', code);
+  parameters.append('redirect_uri', redirectUri);
+  parameters.append('code_verifier', codeVerifier);
+
+  const { data } = await axios.post<TokenSetParameters>(endpoint, parameters, {
```
simeng-li, line 25: TODO: error handle?

simeng-li: feels like we should have a general error handling solution for all client api calls

packages/client/src/grant-token.ts
gao-sun: discussed offline, no harm for checking this schema, we can also prevent some incidents like mistyped URL. i'm good with this. cc @simeng-li

packages/client/src/errors.ts
```diff
@@ -0,0 +1,41 @@
+import { AxiosResponse } from 'axios';
+import { ZodError } from 'zod';
+
+interface OPErrorParmeters {
+  errorMessage?: string;
```
gao-sun, line 5: maybe original `message` is good to go

packages/client/src/errors.ts
```diff
@@ -0,0 +1,41 @@
...
+  errorUri?: string;
+  errorDescription?: string;
```
gao-sun, line 7: discussed offline, description can be extracted from phrases package  like `sdk-phrases` to align error code/message across SDKs

packages/client/src/errors.ts
```diff
@@ -0,0 +1,41 @@
...
+  zodError?: ZodError;
+  originalError?: Error;
+}
+
+export class OPError extends Error {
+  public errorMessage?: string;
+  public errorUri?: string;
+  public errorDescription?: string;
```
gao-sun, line 15: maybe no need for `error` prefix?

packages/client/src/op-request.ts
```diff
@@ -0,0 +1,28 @@
...
+  async (error: Error) => {
+    return Promise.reject(
+      new OPError({
+        errorMessage: 'request error with OP',
+        errorDescription: error.message,
+      })
+    );
+  }
+);
+axiosInstance.interceptors.response.use(
```
gao-sun, line 17:
```suggestion
);

axiosInstance.interceptors.response.use(
```

packages/client/src/op-request.ts
```diff
@@ -0,0 +1,28 @@
+import axios, { AxiosError } from 'axios';
+
+import { OPError } from './errors';
+
+const axiosInstance = axios.create();
+axiosInstance.interceptors.request.use(
```
gao-sun, line 6:
```suggestion
const axiosInstance = axios.create();

axiosInstance.interceptors.request.use(
```

packages/client/src/errors.ts
wangsijie: LOG-173

packages/client/src/op-request.ts
```diff
@@ -0,0 +1,28 @@
...
+        message: `request error with OP (${error.message})`,
+      })
+    );
+  }
+);
+
+axiosInstance.interceptors.response.use(
+  (response) => response,
+  async (error: AxiosError) => {
+    const opError = new OPError({
```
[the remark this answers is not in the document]
wangsijie: [LOG-174](https://linear.app/silverhand/issue/LOG-174/error-handler-%E6%98%AF%E5%90%A6%E5%A4%84%E7%90%86-axios-error)

packages/client/src/errors.ts
```diff
@@ -0,0 +1,30 @@
+import { AxiosResponse } from 'axios';
+import { ZodError } from 'zod';
+
+interface OPErrorParmeters {
+  message?: string;
+  uri?: string;
+  zodError?: ZodError;
+  originalError?: Error;
```
gao-sun, line 8: maybe combine these errors like property `originalError`

packages/client/src/grant-token.ts
```diff
@@ -0,0 +1,40 @@
...
+
+  const response = await opRequest.post(endpoint, parameters, {
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
+    },
+  });
+  const result = TokenSetParametersSchema.safeParse(response.data);
+
+  if (!result.success) {
+    // Will get a lint error without `as ZodError`, wired
```
gao-sun, line 35: typo: weird

gao-sun: we can just disable the rule for this line

PR #43, opened 2021-10-15
feat: requestLogin

wangsijie:

### DIFF
--- a/packages/client/src/request-login.ts
+++ b/packages/client/src/request-login.ts
@@ -1,0 +1,24 @@
+import qs from 'query-string';
+
+import { generateCodeChallenge, generateCodeVerifier } from './generators';
+
+export const getLoginUrlAndCodeVerifier = (
+  baseUrl: string,
+  clientId: string,
+  redirectUri: string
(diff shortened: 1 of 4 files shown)


packages/client/src/request-login.ts
```diff
@@ -0,0 +1,24 @@
...
+    client_id: clientId,
+    scope: 'openid offline_access',
+    response_type: 'code',
+    redirect_uri: [redirectUri],
+    prompt: 'consent',
+    code_challenge: codeChallenge,
+    code_challenge_method: 'S256',
+  })}`;
+
+  return [url, codeVerifier];
```
simeng-li, line 23: Would defining the return type as a map better?  It might contain more informative keys on the return type rather than current [string, string]

simeng-li (approved):

packages/client/src/request-login.ts
gao-sun: or we can write js docs for this func. i'm ok with both

gao-sun (approved):

PR #44, opened 2021-10-15
refactor: use opRequest in discover

wangsijie:
## Summary

Replace axios in method discover with opRequest

### DIFF
--- a/packages/client/src/discover.ts
+++ b/packages/client/src/discover.ts
@@ -1,4 +1,4 @@
-import { z, ZodError } from 'zod';
+import { z } from 'zod';
 
 import { OPError } from './errors';
 import { opRequest } from './op-request';
@@ -28,7 +28,8 @@ export default async function discover(url: string): Promise<OIDCConfiguration> 
   const result = OIDCConfigurationSchema.safeParse(response.data);
 
   if (!result.success) {
-    throw new OPError({ originalError: result.error as ZodError });
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+    throw new OPError({ originalError: result.error });
   }
 
   return result.data;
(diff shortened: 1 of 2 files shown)


packages/client/src/discover.ts
```diff
@@ -20,8 +22,14 @@ const appendSlashIfNeeded = (url: string): string => {
...
 export default async function discover(url: string): Promise<OIDCConfiguration> {
-  const { data } = await axios.get<OIDCConfiguration>(
+  const response = await opRequest.get(
     `${appendSlashIfNeeded(url)}oidc/.well-known/openid-configuration`
   );
-  return OIDCConfigurationSchema.parse(data);
+  const result = OIDCConfigurationSchema.safeParse(response.data);
+
+  if (!result.success) {
+    throw new OPError({ originalError: result.error as ZodError });
```
gao-sun, line 31: let's disable the rule instead of using `as`

gao-sun (changes requested):

packages/client/src/grant-token.ts
```diff
@@ -32,8 +32,7 @@ export const grantTokenByAuthorizationCode = async (
   const result = TokenSetParametersSchema.safeParse(response.data);
 
   if (!result.success) {
-    // Will get a lint error without `as ZodError`, weird
-    throw new OPError({ originalError: result.error as ZodError });
+    throw new OPError({ originalError: result.error });
```
gao-sun, line 35: let's add a linear issue of renaming to `LogtoError`

gao-sun (approved):

packages/client/src/grant-token.ts
wangsijie: LOG-176

PR #45, opened 2021-10-15
feat: grantTokenByRefreshToken

wangsijie:
## Summary

grant a new tokenSet from OP by a refresh_token

## Testing

call with a real refresh token in `logto.dev`

image

### DIFF
--- a/packages/client/src/grant-token.test.ts
+++ b/packages/client/src/grant-token.test.ts
@@ -1,6 +1,6 @@
 import nock from 'nock';
 
-import { grantTokenByAuthorizationCode } from './grant-token';
+import { grantTokenByAuthorizationCode, grantTokenByRefreshToken } from './grant-token';
 
 describe('grantTokenByAuthorizationCode', () => {
   test('get tokenSet paramaters', async () => {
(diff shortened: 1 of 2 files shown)


packages/client/src/grant-token.ts
```diff
@@ -38,3 +38,27 @@ export const grantTokenByAuthorizationCode = async (
 
   return result.data;
 };
+
+export const grantTokenByRefreshToken = async (
+  endpoint: string,
+  clientId: string,
+  refreshToken: string
```
simeng-li, line 45: just curious without the PKCE validation, could it be malicious attack base a refresh_token grant flow?

simeng-li (approved):
LGTM

packages/client/src/grant-token.test.ts
```diff
@@ -46,3 +46,26 @@ describe('grantTokenByAuthorizationCode', () => {
...
+describe('grantTokenByRefreshToken', () => {
+  test('get tokenSet paramaters', async () => {
+    const successResponse = {
+      access_token: 'access_token',
+      expires_in: 3600,
+      id_token: 'id_token',
+      refresh_token: 'refresh_token',
+    };
+
+    nock('https://logto.dev', { allowUnmocked: true })
```
gao-sun, line 59: The URL should be a constant? Saw it several times.

PR #50, opened 2021-10-19
chore: webpack

wangsijie:
## Summary

use webpack to package client into a single bundle file, this can be useful in vanilla js.


simeng-li (approved):
LGTM

PR #65, opened 2021-10-29
feat: getAccessToken

wangsijie:
## Summary

method #getAccessToken

Return accessToken from tokenSet if is not expired, otherwise, use refreshToken to grant new accessToken

## Linear Issue Reference

(LOG-221)

## Testing

unit tests


simeng-li (approved):
LGTM

PR #69, opened 2021-11-01
fix(playground): migrate playground to js repo

simeng-li:
## Summary

migrate playground to js repo as a monorepo

logto-io/playground -> logto-io/js/packages/playground

## Testing
test on local server

@wangsijie cc: @gao-sun @IceHe @xiaoyijun


wangsijie (approved):
LGTM

PR #71, opened 2021-11-02
feat: custom scope

IceHe:
## Summary

- The developers can use custom `scope` value.
- LogtoClient can correct the value of `scope` from parameters.

### DIFF
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -15,9 +15,10 @@ jest.mock('./verify-id-token');
 const DOMAIN = 'logto.dev';
 const BASE_URL = `https://${DOMAIN}`;
 const CLIENT_ID = 'client1';
+const DEFAULT_SCOPE = 'openid offline_access';
 const SUBJECT = 'subject1';
 const REDIRECT_URI = 'http://localhost:3000';
-const SESSEION_MANAGER_KEY = 'LOGTO_SESSION_MANAGER';
+const SESSION_MANAGER_KEY = 'LOGTO_SESSION_MANAGER';
 const REDIRECT_CALLBACK = `${REDIRECT_URI}?code=authorization_code`;
 const REDIRECT_CALLBACK_WITH_ERROR = `${REDIRECT_CALLBACK}&error=invalid_request&error_description=code_challenge%20must%20be%20a%20string%20with%20a%20minimum%20length%20of%2043%20characters`;
 
@@ -111,10 +112,13 @@ describe('LogtoClient', () => {
 
     test('claims restored', async () => {
       const storage = new MemoryStorage();
-      storage.setItem(`LOGTO_TOKEN_SET_CACHE::${discoverResponse.issuer}::${CLIENT_ID}`, {
-        ...fakeTokenResponse,
-        id_token: (await generateIdToken()).idToken,
-      });
+      storage.setItem(
+        `LOGTO_TOKEN_SET_CACHE::${discoverResponse.issuer}::${CLIENT_ID}::${DEFAULT_SCOPE}`,
+        {
+          ...fakeTokenResponse,
+          id_token: (await generateIdToken()).idToken,
+        }
+      );
       const logto = await LogtoClient.create({
         domain: DOMAIN,
         clientId: CLIENT_ID,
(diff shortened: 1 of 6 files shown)


packages/client/src/index.ts
```diff
@@ -22,6 +23,7 @@ export interface ConfigParameters {
   domain: string;
   clientId: string;
   storage?: ClientStorage;
+  scope: string | string[];
```
wangsijie, line 26: can be optional

packages/client/src/index.ts
```diff
@@ -35,27 +37,53 @@ export const appendSlashIfNeeded = (url: string): string => {
...
+    const { clientId, storage, scope } = config;
     this.clientId = clientId;
+    this.scope = LogtoClient.correctScope(scope);
     this.oidcConfiguration = oidcConfiguration;
     this.storage = storage ?? new LocalStorage();
     this.sessionManager = new SessionManager(this.storage);
     this.createTokenSetFromCache();
   }
 
+  static correctScope(originalScope: string | string[]): string {
```
wangsijie, line 54: put it into utils? no need to be public

packages/client/src/index.ts
```diff
@@ -35,27 +37,53 @@ export const appendSlashIfNeeded = (url: string): string => {
...
+      }
+
+      return OPENID + ' ' + originalScope.map((s) => s.trim()).join(' ');
+    }
+
+    if (originalScope.trim().length === 0) {
+      return OPENID;
+    }
+
+    if (originalScope.split(' ').includes(OPENID)) {
```
wangsijie, line 71: this `includes` code is seen twice,
how about turning string value into `array` and then check default scopes,
also, `offline_access` is necessary as well

wangsijie (changes requested):

packages/client/src/index.ts
```diff
@@ -35,27 +37,53 @@ export const appendSlashIfNeeded = (url: string): string => {
...
-    const client = new LogtoClient(config, await discover(`https://${config.domain}`));
-    return client;
+    return new LogtoClient(config, await discover(`https://${config.domain}`));
   }
 
   public async loginWithRedirect(redirectUri: string) {
     const { url, codeVerifier } = await getLoginUrlAndCodeVerifier(
       this.oidcConfiguration.authorization_endpoint,
       this.clientId,
+      this.scope,
```
simeng-li, line 86: there are too many input attributes here, better make it a single payload map

packages/client/src/index.ts
simeng-li: could align with auth0 only accepts string with empty space split the scopes

packages/client/src/index.test.ts
```diff
@@ -16,9 +16,11 @@ jest.mock('./verify-id-token');
 const DOMAIN = 'logto.dev';
 const BASE_URL = `https://${DOMAIN}`;
 const CLIENT_ID = 'client1';
+const OPENID = 'openid';
+const SCOPE = OPENID;
```
simeng-li, line 20: feels like no need to define a const SCOPE = OPENID here.  We could define the value whenever it is being used. i.e. new LogtoClient
as scope should be optional param.

1. be default no need to pass in Scope to all the current existing tests. 
2. Create a test does not pass in Scope as check the `client.scope` contains OPENID
3. Create a test pass scope = OPENID, check the `client.scope` contains only one single OPENID
4. Create a test pass scope = `${OPENID} email`, check the `client.scope` contains OPENID and email

should be good enough

packages/client/src/index.ts
simeng-li: can we could have the trim check on the original scope input param

IceHe: i prefer `string | string[]`

wangsijie (changes requested):
`tokenSetCacheKey` should include `scope`

wangsijie (approved):

packages/client/src/index.ts
```diff
@@ -149,7 +153,7 @@ export default class LogtoClient {
   }
 
   private get tokenSetCacheKey() {
-    return `${TOKEN_SET_CACHE_KEY}::${this.oidcConfiguration.issuer}::${this.clientId}`;
+    return `${TOKEN_SET_CACHE_KEY}::${this.oidcConfiguration.issuer}::${this.clientId}::${this.scope}`;
```
simeng-li, line 156: @IceHe create a issue replace spaces in scope as it is being used as party of the key.

packages/client/src/utils.ts
```diff
@@ -57,3 +57,19 @@ export const nowRoundToSec = () => Math.floor(Date.now() / 1000);
...
+ */
+export const generateScope = (originalScope?: string | string[]): string => {
+  const originalScopeValues =
+    originalScope === undefined
+      ? []
+      : Array.isArray(originalScope)
+      ? originalScope
+      : originalScope.split(' ');
+  const nonEmptyScopeValues = originalScopeValues.map((s) => s.trim()).filter((s) => s.length > 0);
+  const uniqueScopeValues = new Set(['openid', 'offline_access', ...nonEmptyScopeValues]);
```
simeng-li, line 73: better create a const for default scopes

PR #72, opened 2021-11-02
chore: upgrade jose to 4.x

wangsijie:
## Summary

in jose 3.x, module can only be imported by subpath, it could cause problems in both jsdom and webpack.

change log from jose 4.x:

> All module named exports have moved from subpaths to just "jose". For example, import { jwtVerify } from 'jose/jwt/verify' is now just import { jwtVerify } from 'jose'.

## Testing

old unit tests.


simeng-li (approved):
LGTM

PR #74, opened 2021-11-02
feat(react): init provider & hook

wangsijie:
## Summary

init package `@logto/react`

## Testing

a simple unit test is added.

### DIFF
--- a/packages/react/src/use-logto.spec.tsx
+++ b/packages/react/src/use-logto.spec.tsx
@@ -7,23 +7,23 @@ import useLogto from './use-logto';
 const isAuthenticated = jest.fn();
 
 jest.mock('@logto/client', () => {
-  const Mocked = jest
-    .fn()
-    .mockImplementation(({ onAuthStateChange }: { onAuthStateChange: () => void }) => {
-      return {
-        isAuthenticated,
-        handleCallback: jest.fn().mockImplementation(() => {
-          onAuthStateChange();
-        }),
-        logout: jest.fn().mockImplementation(() => {
-          onAuthStateChange();
-        }),
-      };
-    });
+  const Mocked = jest.fn(({ onAuthStateChange }: { onAuthStateChange: () => void }) => {
+    return {
+      isAuthenticated,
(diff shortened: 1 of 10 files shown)


packages/react/src/use-logto.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+import React from 'react';
+
+import { LogtoProvider } from './provider';
+import useLogto from './use-logto';
+
+const isAuthenticated = jest.fn();
+
+jest.mock('@logto/client', () => {
+  const Mocked = jest
+    .fn()
```
simeng-li, line 11: nit: could just use
```js
jest.fn(({ onAuthStateChange }: { onAuthStateChange: () => void }) => ({
  isAuthenticated,
   handleCallback: jest.fn().mockImplementation(() => {
      onAuthStateChange();
   }),
   logout: jest.fn().mockImplementation(() => {
      onAuthStateChange();
    }),
}));
```

packages/react/src/use-logto.spec.tsx
```diff
@@ -0,0 +1,109 @@
...
+        }),
+        logout: jest.fn().mockImplementation(() => {
+          onAuthStateChange();
+        }),
+      };
+    });
+  return {
+    default: Mocked,
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    create: jest.fn().mockImplementation((...args) => new Mocked(...args)),
```
simeng-li, line 26: ditto: `jest.fn((...args) => new Mocked(...args)),`

packages/react/src/use-logto.ts
```diff
@@ -0,0 +1,21 @@
+import { useContext } from 'react';
+
+import { LogtoContext, LogtoContextProperties } from './context';
+
+export default function useLogto() {
+  const context = useContext<LogtoContextProperties>(LogtoContext);
```
simeng-li, line 6: nit: no need to specify the return types here. useContext should recognize the return type automatically.

simeng-li (approved):
overall LGTM

PR #77, opened 2021-11-03
feat: isLoginRedirect

wangsijie:
## Summary

A util to check if current webpage state is `login callback`

## Testing

`isLoginRedirect` in `index.test.ts`


simeng-li (approved):
LGTM

PR #78, opened 2021-11-03
refactor: generators.ts

IceHe:
## Summary

I got some new understanding of code verifier after re-reading https://datatracker.ietf.org/doc/html/rfc7636#section-4.1
So correct some code.


simeng-li (approved):
LGTM

PR #80, opened 2021-11-03
refactor: extract constants

IceHe:
## Summary

Gather constants into a file -- `constants.ts`

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -1,5 +1,6 @@
 import { Optional } from '@silverhand/essentials';
 
+import { TOKEN_SET_CACHE_KEY } from './constants';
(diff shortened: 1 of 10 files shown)


packages/client/src/index.ts
```diff
@@ -185,3 +184,4 @@ export default class LogtoClient {
     }
   }
 }
+export { STORAGE_KEY_PREFIX } from './constants';
```
IceHe, line 187: TBD: SHOULD export again?

wangsijie (approved):

packages/client/src/index.ts
wangsijie: no need for that, better be private

PR #81, opened 2021-11-03
refactor(playground): integrate playground with client js sdk

simeng-li:
## Summary

Integrate playground with client js sdk

- Create a logtoClient instance and put is in as a context
- replace original login/logout logic using logtoClient api
- add `jest-setup` file to bypass the `jose` jest env issue
- As `token_set` storage is hidden to the client, display IDToken claims instead of original token_set details  after success login

## Testing
Test locally successfully sign-in and out

image

@wangsijie @xiaoyijun cc: @IceHe @gao-sun

### DIFF
--- a/packages/playground/src/consts/url.ts
+++ b/packages/playground/src/consts/url.ts
@@ -1,5 +1,2 @@
 export const baseUrl = `http://${String(process.env.HOST)}:${String(process.env.PORT)}`;
 export const oidcDomain = process.env.RAZZLE_OIDC_BASE_DOMAIN ?? 'N/A';
-export const oidcUrl = process.env.RAZZLE_OIDC_BASE_DOMAIN
-  ? `http://${process.env.RAZZLE_OIDC_BASE_DOMAIN}/oidc`
-  : 'N/A';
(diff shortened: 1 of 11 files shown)


xiaoyijun (approved):
LGTM

packages/playground/src/consts/url.ts
```diff
@@ -1,2 +1,5 @@
 export const baseUrl = `http://${String(process.env.HOST)}:${String(process.env.PORT)}`;
-export const oidcUrl = process.env.RAZZLE_OIDC_BASE_URL ?? 'N/A';
+export const oidcDomain = process.env.RAZZLE_OIDC_BASE_DOMAIN ?? 'N/A';
+export const oidcUrl = process.env.RAZZLE_OIDC_BASE_DOMAIN
```
wangsijie, line 3: is this still useful?

wangsijie (approved):
LGTM

packages/playground/src/consts/url.ts
simeng-li: removed thx

PR #82, opened 2021-11-04
feat: onAuthStateChange

wangsijie:
## Summary

Add support for an optional listener on authenticated state change

## Testing

add test cases in `index.test.ts`

### DIFF
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -398,7 +398,7 @@ describe('LogtoClient', () => {
         storage,
         onAuthStateChange,
       });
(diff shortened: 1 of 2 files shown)


IceHe:
LGTM

IceHe (approved):

simeng-li (commented):
open to discuss: One option is simply add a callback method, and let the consumer to decide whether call the `client.isAuthed()` to get the latest authenticated status?  
Or one more option maybe instead of return the `auth` param hard code to `true` or `false` could return `this.isAuth()`?

PR #83, opened 2021-11-04
test: import constants.ts

IceHe:
## Summary

Extracted common constants to `constants.ts` and forgot to replacing related constants in `*.test.ts`.
So replace common constants in `*.test.ts` now.

## Testing

Pass all existing unit tests.

### DIFF
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -7,6 +7,7 @@ import { SignJWT, generateKeyPair } from 'jose';
 import nock from 'nock';
 
 import LogtoClient from '.';
+import { DEFAULT_SCOPE_STRING, SESSION_MANAGER_KEY } from './constants';
 import { MemoryStorage } from './storage';
 import { verifyIdToken } from './verify-id-token';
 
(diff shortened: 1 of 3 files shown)


packages/client/src/index.test.ts
```diff
@@ -113,7 +112,7 @@ describe('LogtoClient', () => {
     test('claims restored', async () => {
       const storage = new MemoryStorage();
       storage.setItem(
-        `LOGTO_TOKEN_SET_CACHE::${discoverResponse.issuer}::${CLIENT_ID}::${DEFAULT_SCOPE}`,
+        `LOGTO_TOKEN_SET_CACHE::${discoverResponse.issuer}::${CLIENT_ID}::${DEFAULT_SCOPE_STRING}`,
```
gao-sun, line 115: the string itself should be a const as well?

gao-sun (changes requested):
please fulfill PR summary and testing part

packages/client/src/index.test.ts
IceHe: - All string constants:
    The constants looks like symbol links.
    When we need to modify its value, we can modify its value at the correct position (where symbol link referenced to) and then rename it.
    It's more accurate than replacing all strings via searching because the variables with the same values maybe have the different meanings.
- `DEFAULT_SCOPE_STRING`:
    So far it contains 'openid' and 'offline_access'.
    Maybe it will contain more scope values.
    Extract it to a constant so that we can make it easier to modify.

packages/client/src/index.test.ts
```diff
@@ -146,6 +144,7 @@ describe('LogtoClient', () => {
 
   describe('loginWithRedirect', () => {
     test('window.location.assign should have been called', async () => {
+      jest.fn();
```
simeng-li, line 147: typo?

IceHe: My bad.

packages/client/src/index.test.ts
```diff
@@ -7,24 +7,25 @@ import { SignJWT, generateKeyPair } from 'jose';
...
 const DOMAIN = 'logto.dev';
 const BASE_URL = `https://${DOMAIN}`;
+const ISSUER = `${BASE_URL}/oidc`;
 const CLIENT_ID = 'client1';
-const DEFAULT_SCOPE = 'openid offline_access';
 const SUBJECT = 'subject1';
 const REDIRECT_URI = 'http://localhost:3000';
-const SESSION_MANAGER_KEY = 'LOGTO_SESSION_MANAGER';
 const REDIRECT_CALLBACK = `${REDIRECT_URI}?code=authorization_code`;
 const REDIRECT_CALLBACK_WITH_ERROR = `${REDIRECT_CALLBACK}&error=invalid_request&error_description=code_challenge%20must%20be%20a%20string%20with%20a%20minimum%20length%20of%2043%20characters`;
```
gao-sun, line 23: should use some utils to generate this. can add a linear issue and do it later

gao-sun (approved):

PR #84, opened 2021-11-04
feat: protected route

wangsijie:
## Summary

`<ProtectedRoute path="/">xxx</ProtectedRoute>`

When user enters secure route (protected route), will trigger login process automatically, ensure that only see content in login state.

## Linear Issue Reference

[LOG-231](https://linear.app/silverhand/issue/LOG-231/react-react-router)

## Testing

`ProtectedRoute.spec.ts`

### DIFF
--- a/packages/react/src/ProtectedRoute.spec.tsx
+++ b/packages/react/src/ProtectedRoute.spec.tsx
@@ -1,0 +1,95 @@
+import { render, waitFor, screen } from '@testing-library/react';
+import React from 'react';
+import { MemoryRouter } from 'react-router-dom';
+
+import { LogtoProvider } from '.';
+import ProtectedRoute from './ProtectedRoute';
+
+const loginWithRedirect = jest.fn();
+const isAuthenticated = jest.fn();
+
+jest.mock('@logto/client', () => {
+  const Mocked = jest.fn().mockImplementation(() => {
+    return {
+      loginWithRedirect,
+      isAuthenticated,
+    };
+  });
+  return {
+    default: Mocked,
+    // eslint-disable-next-line @typescript-eslint/no-unsafe-return
+    create: jest.fn().mockImplementation(() => new Mocked()),
+  };
+});
+
+const renderWrapper = (children: React.ReactNode) => {
+  return (
+    <MemoryRouter>
+      <LogtoProvider
+        logtoConfig={{
+          domain: 'logto.dev',
+          clientId: 'client',
+        }}
+      >
+        {children}
+      </LogtoProvider>
+    </MemoryRouter>
+  );
+};
+
+afterEach(() => {
+  isAuthenticated.mockClear();
+  loginWithRedirect.mockClear();
+});
+
+describe('ProtectedRoute', () => {
+  describe('not authenticated', () => {
+    beforeEach(() => {
+      isAuthenticated.mockImplementation(() => false);
+    });
+
+    test('loginWithRedirect should be called once', async () => {
+      render(renderWrapper(<ProtectedRoute path="/" />));
+
+      await waitFor(() => {
+        expect(loginWithRedirect).toHaveBeenCalled();
+      });
+
+      expect(loginWithRedirect).toHaveBeenCalledTimes(1);
+    });
+
+    test('protected content should not be rendered', async () => {
+      render(renderWrapper(<ProtectedRoute path="/">protected</ProtectedRoute>));
+
+      await waitFor(() => {
+        expect(loginWithRedirect).toHaveBeenCalled();
+      });
+
+      expect(screen.queryByText('protected')).toBeNull();
+    });
+  });
+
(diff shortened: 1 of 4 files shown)


packages/react/src/provider.tsx
```diff
@@ -20,6 +20,8 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
           setIsAuthenticated(client.isAuthenticated());
         },
       });
+      setIsAuthenticated(client.isAuthenticated());
+      // Help me: need to ensure that executes after `setIsAuthenticated`
       setLogtoClient(client);
```
gao-sun, line 25: didn't read the context, but intuitively, either use 2 `useEffect()` (less pretty) either re-design the data flow. @simeng-li for more context

packages/react/src/SecureRoute.tsx
```diff
@@ -0,0 +1,45 @@
...
+import { Route, RouteProps } from 'react-router-dom';
+
+import useLogto from './use-logto';
+
+const SecureRoute: FC<RouteProps> = ({ ...routeProperties }) => {
+  const { isAuthenticated, logtoClient, isLoading } = useLogto();
+  // The following useEffect takes several deps,
+  // make sure only happen when isAuthenticated is changed,
+  // prevent calling multiple times
+  const pendingLoginReference = useRef(false);
```
gao-sun, line 11: don't know if `useRef()` here is the right choice. it's more like a state from the code below.

gao-sun: it's okay to call the `useEffect` multiple times, as long as we can ensure the right one (with `login()` called) will be just called once when all criteria are satisfied.

packages/react/src/SecureRoute.tsx
```diff
@@ -0,0 +1,45 @@
...
+  useEffect(() => {
+    const login = async () => {
+      if (pendingLoginReference.current) {
+        return;
+      }
+
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      pendingLoginReference.current = true;
+
+      await logtoClient!.loginWithRedirect(window.location.href);
```
gao-sun, line 22: let's ban `foo!` in general. we can define the function after we check `logtoClient`

gao-sun (changes requested):

packages/react/src/SecureRoute.tsx
wangsijie: use `pendingLoginReference` as a `lock` for async function `login`, to ensure there is only one excecuting `login`, why useRef instead of useState?
1. will NOT cause re-render or call `useEffect`, avoid loop.
2. not be used in `render`

wangsijie: outdated, thouse code has been removed

packages/react/src/provider.tsx
wangsijie: outdated, thouse code has been removed

packages/react/src/ProtectedRoute.tsx
```diff
@@ -0,0 +1,24 @@
...
+  useEffect(() => {
+    if (!isInitialized || isLoading || isAuthenticated) {
+      return;
+    }
+
+    void loginWithRedirect(window.location.href);
+  }, [isAuthenticated, isInitialized, loginWithRedirect]);
+
+  if (!isAuthenticated) {
+    return null;
```
simeng-li, line 18: Do we intend to leave the developers to handler their own loading status apart from the `ProtectedRoute`?

packages/react/src/ProtectedRoute.spec.tsx
```diff
@@ -0,0 +1,95 @@
...
+
+      await waitFor(() => {
+        expect(loginWithRedirect).toHaveBeenCalled();
+      });
+
+      expect(loginWithRedirect).toHaveBeenCalledTimes(1);
+    });
+
+    test('protected content should not be rendered', async () => {
+      render(renderWrapper(<ProtectedRoute path="/">protected</ProtectedRoute>));
```
simeng-li, line 62: nit: just FYI could `render` methods returns multiple query api itself

```js
const {  queryByText } = render(renderWrapper(<ProtectedRoute path="/">protected</ProtectedRoute>));
```

instead of the whole screen, this returned queryByText already includes the rendered ReactDom context.

simeng-li (approved):
overall LGTM

packages/react/src/ProtectedRoute.tsx
wangsijie: Yes, we just export `isInitialized` and `isLoading`

packages/react/src/ProtectedRoute.spec.tsx
wangsijie: Thanks

gao-sun (approved):

PR #85, opened 2021-11-05
refactor(js): generators.ts and utils.ts

IceHe:
## Summary

- generators.ts
    - Package all existing exports into `Generators`
    - Export default `Generators`
- utils.ts
    - Fix method `UrlSafeBase64.encode` (originaly `encodeBase64`)
    - Fix method `UrlSafeBase64.decode` (originaly `decodeBase64`)

## Testing

- Pass all existing unit tests.
- Assert `rawContent = decode(encode(rawContent))`

### DIFF
--- a/packages/client/src/generators.ts
+++ b/packages/client/src/generators.ts
@@ -7,7 +7,6 @@ import {
   CODE_VERIFIER_ALPHABET,
   CODE_VERIFIER_MAX_LENGTH,
   DEFAULT_SCOPE_VALUES,
-  RANDOM_STRING_MAX_LENGTH,
 } from './constants';
 import { UrlSafeBase64 } from './utils';
 
@@ -20,20 +19,13 @@ function generateRandomString(length: number) {
 
 /**
  * Generates random bytes and encodes them in url safe base64.
- * @param length The length of the string (before base64url-encoding) to generate.
  */
-export const generateRandomInUrlSafeBase64 = (length = RANDOM_STRING_MAX_LENGTH) =>
-  UrlSafeBase64.encode(generateRandomString(length));
+export const generateState = () => generateRandomString(CODE_VERIFIER_MAX_LENGTH);
 
 /**
  * Generates random bytes and encodes them in url safe base64.
  */
-export const generateState = () => generateRandomInUrlSafeBase64();
-
-/**
- * Generates random bytes and encodes them in url safe base64.
- */
-export const generateNonce = () => generateRandomInUrlSafeBase64();
+export const generateNonce = () => generateRandomString(CODE_VERIFIER_MAX_LENGTH);
 
 /**
  * Generates code verifier use max_length
--- a/packages/client/src/generators.ts
+++ b/packages/client/src/generators.ts
@@ -9,7 +9,7 @@ import {
   DEFAULT_SCOPE_VALUES,
   RANDOM_STRING_MAX_LENGTH,
 } from './constants';
-import { encodeBase64 } from './utils';
+import { UrlSafeBase64 } from './utils';
 
 /**
  * @param length The length of the string to generate.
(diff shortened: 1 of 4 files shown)


packages/client/src/utils.ts
```diff
@@ -57,8 +57,21 @@ export const decodeToken = (token: string): IDToken => {
 
 export const nowRoundToSec = () => Math.floor(Date.now() / 1000);
 
-export const encodeBase64 = (input: string) => btoa(input);
-export const decodeBase64 = (input: string) => atob(input);
+const replaceNonUrlSafeCharactersInBase64 = (rawString: string) => {
```
simeng-li, line 60: https://github.com/RGBboy/urlsafe-base64/blob/master/lib/urlsafe-base64.js looks pretty handy

packages/client/src/generators.ts
```diff
@@ -21,18 +21,18 @@ function generateRandomString(length: number) {
  * Generates random bytes and encodes them in url safe base64.
  * @param length The length of the string (before base64url-encoding) to generate.
  */
-export const generateRandomInBase64 = (length = RANDOM_STRING_MAX_LENGTH) =>
-  encodeBase64(generateRandomString(length));
+export const generateRandomInUrlSafeBase64 = (length = RANDOM_STRING_MAX_LENGTH) =>
+  encodeInUrlSafeBase64(generateRandomString(length));
```
simeng-li, line 25: just `encodeUrlSafeBase64` will do

packages/client/src/utils.ts
simeng-li: or we could fork the usage of this package. look pretty elegant to me 

```js
const UrlSafeBase64 = {
 encode: () => {},
 decode: () => {},
}
```

packages/client/src/generators.ts
simeng-li: also is there a standard for us, which method should lies under `utils` which should remain in this file? looks to me `generateRandomString` & `encodeUrlSafeBase64` share a same use case. 
- both are only used in this file
- both are general string operators

either move all to `utils` or keep in `generators` works for me . Just need to align a standard

IceHe: - I prefer to put all `generate*()` methods in `generators.ts`.

    I will move `generateScope()` from `utils.ts` to `generators.ts` in another PR.

wangsijie (approved):

packages/client/src/utils.ts
```diff
@@ -55,5 +55,19 @@ export const decodeToken = (token: string): IDToken => {
...
-export const decodeBase64 = (input: string) => atob(input);
+const replaceNonUrlSafeCharacters = (base64String: string) =>
+  base64String
+    .replace(/\+/g, '-') // Convert '+' to '-'
+    .replace(/\//g, '_') // Convert '/' to '_'
+    .replace(/=+$/g, ''); // Remove ending '='
+const restoreNonUrlSafeCharacters = (base64String: string) =>
+  base64String
+    .replace(/-/g, '+') // Convert '-' to '+'
+    .replace(/_/g, '/'); // Convert '_' to '/'
```
gao-sun, line 66: nit: the comments is redundant, see [this doc](https://www.cs.utah.edu/~germain/PPS/Topics/commenting.html#:~:text=How%20to%20comment%20Code%3A,not%20%22self%2Ddocumenting%22.).

gao-sun (changes requested):

packages/client/src/utils.ts
IceHe: The comments are just for who is unfamiliar with regex.
But I agree with you now, there is little need to explain regex to them.

simeng-li (approved):
LGTM

packages/client/src/utils.ts
```diff
@@ -55,5 +55,19 @@ export const decodeToken = (token: string): IDToken => {
...
+
+export const UrlSafeBase64 = {
+  encode: (rawString: string) => replaceNonUrlSafeCharacters(btoa(rawString)),
+  decode: (encodedString: string) => atob(restoreNonUrlSafeCharacters(encodedString)),
+  replaceNonUrlSafeCharacters,
```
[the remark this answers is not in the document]
IceHe: - `replaceNonUrlSafeCharacters ` is also used by `generateCodeChallenge` method in `utils.ts`, so I decided to export it for reuse.
- `restoreNonUrlSafeCharacters` hasn't been used, so I can remove from `UrlSafeBase64`.

TBD @simeng-li :

- Move `replaceNonUrlSafeCharacters ` method out of `UrlSafeBase64`?
- DO NOT export `restoreNonUrlSafeCharacters` method?
- More suggestions?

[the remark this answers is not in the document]
gao-sun: we can move them to our [essentials](https://github.com/silverhand-io/essentials) package.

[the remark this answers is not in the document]
gao-sun: let's add a linear issue before merging.

gao-sun (approved):

packages/client/src/utils.ts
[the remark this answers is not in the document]
IceHe: done.

PR #86, opened 2021-11-05
docs(playground): update readme

simeng-li:
## Summary

Update the playground readme file
as we integrate the playground with `logtoClient` which take OIDC domain instead of the full url as its input config.  

## Testing
Readme only no need to test


xiaoyijun (approved):
LGTM

PR #92, opened 2021-11-09
feat(react): auto handle callback

wangsijie:
## Summary

use `logtoClient.isLoginCallback` to auto detect if need to handle callback

## Linear Issue Reference

[LOG-232](https://linear.app/silverhand/issue/LOG-232/react-logincallback-comp)

## Testing

`use-logto.spec.tsx`

### DIFF
--- a/packages/react/src/provider.tsx
+++ b/packages/react/src/provider.tsx
@@ -18,8 +18,14 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
     const createClient = async () => {
       try {
         const client = await LogtoClient.create(logtoConfig);
-        dispatch({ type: 'INITIALIZE', isAuthenticated: client.isAuthenticated() });
         setLogtoClient(client);
+        const isLoginRedirect = client.isLoginRedirect(window.location.href);
+        // If is login redirect, should set isLoading and isInitialized at the same time
+        dispatch({
+          type: 'INITIALIZE',
+          isAuthenticated: client.isAuthenticated(),
+          isLoading: isLoginRedirect,
+        });
       } catch (error: unknown) {
         dispatch({ type: 'ERROR', error });
       }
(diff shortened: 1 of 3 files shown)


packages/react/src/provider.tsx
```diff
@@ -18,8 +18,14 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
     const createClient = async () => {
       try {
         const client = await LogtoClient.create(logtoConfig);
-        dispatch({ type: 'INITIALIZE', isAuthenticated: client.isAuthenticated() });
         setLogtoClient(client);
+        const isLoginRedirect = client.isLoginRedirect(window.location.href);
+        // If is login redirect, should set isLoading and isInitialized at the same time
+        dispatch({
+          type: 'INITIALIZE',
+          isAuthenticated: client.isAuthenticated(),
```
simeng-li, line 26: usually all dispatch data are wrapped up using a payload object
```js
dispatch({
  type: 'INITIALIZE',
  payload: {
    isAuthenticated,
    isLoading,
  }
})
```

so on the reducer method 

```js
(state, action: { type,  payload }) => { ... }
```

packages/react/src/provider.tsx
```diff
@@ -18,8 +18,14 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
...
+          isLoading: isLoginRedirect,
+        });
       } catch (error: unknown) {
         dispatch({ type: 'ERROR', error });
```
simeng-li, line 30: can error passed into the dispatch with an unknown type?
```js
if (error instance of Error) {
 dispatch({type: 'ERROR', error});
 return
}
throw error;
```

packages/react/src/reducer.ts
```diff
@@ -10,7 +10,12 @@ type Action =
 
 export const reducer = (state: AuthState, action: Action): AuthState => {
   if (action.type === 'INITIALIZE') {
-    return { ...state, isInitialized: true, isAuthenticated: action.isAuthenticated };
+    return {
+      ...state,
+      isInitialized: true,
+      isAuthenticated: action.isAuthenticated,
```
simeng-li, line 16: normally is `action.payload.isAuthenticated`

packages/react/src/provider.tsx
wangsijie: I put this code inside reducer to avoid writing same code everytime before dispatch
```ts
if (action.type === 'ERROR') {
    const { error } = action.payload;
    if (!(error instanceof Error)) {
      throw error;
    }

    return { ...state, error, isLoading: false };
  }
```

simeng-li (approved):
LGTM

PR #93, opened 2021-11-09
fix(client): fix superstruct validation rule

simeng-li:
## Summary

We have encountered a `superstruct` type thrown error specifying a unmatched type of `OIDCConfig`

According to the `struct` official [document](https://docs.superstructjs.org/api-reference/types) 

image

replace `object` declaration using [type](https://docs.superstructjs.org/api-reference/types#type)

## Testing
test locally on playground

@wangsijie cc: @IceHe @xiaoyijun @gao-sun


gao-sun (approved):
lgtm

PR #96, opened 2021-11-10
feat(client): support node

wangsijie:
## Summary

Do 2 things to support running both in node and browser(already supported):

1. can pass in custom `fetchFunction` in constructer
2. onRedirect parameter, defaults to `window.location.assign`

## Testing

change `testEnvironment` to `node` and run test for file `index.test.ts`

### DIFF
--- a/packages/client/src/api.ts
+++ b/packages/client/src/api.ts
@@ -1,3 +1,5 @@
+import { isNode } from '@silverhand/essentials';
+
 import { LogtoError } from './errors';
 
 declare interface LogtoErrorResponse {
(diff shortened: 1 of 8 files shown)


packages/client/src/api.ts
```diff
@@ -18,15 +18,19 @@ const getResponseErrorMessage = async (response: Response): Promise<string> => {
...
 
-export const requestWithFetch = async <T>(...args: Parameters<typeof fetch>): Promise<T> => {
-  const response = await fetch(...args);
-  if (!response.ok) {
-    throw new LogtoError({
-      message: await getResponseErrorMessage(response),
-      response,
-    });
-  }
+export const createRequester = (fetchFunction: typeof fetch = fetch) => {
```
gao-sun, line 21: Should we detect `isNode` here? Looks like user needs to provide a fetch function when use it on NodeJS.

simeng-li (approved):
overall LGTM. one thing cross my mind, since requester is widely used cross the SDK. should we use a  `webpack.ProvidePlugin` to mark the logto requester. so no need to import in everyfile. 

Also just wondering, if developer provide their own requester then for the error thrown logics would it be in-consistent?

wangsijie:
> overall LGTM. one thing cross my mind, since requester is widely used cross the SDK. should we use a `webpack.ProvidePlugin` to alias the logto requester. so no need to import in every file.
> 
> Also just wondering, if developer provide their own requester， then for the error thrown logics， would it be in-consistent?

LOG-262

PR #98, opened 2021-11-11
test(react): mock isLoginRedirect to fix test

wangsijie:
## Summary

Since #92 , `isLoginRedirect` will be called automaticly, but is not mocked in `ProtectedRoute.spec.tsx`, causing test failed.


simeng-li (approved):
LGTM

PR #99, opened 2021-11-15
refactor(client): simplify parseRedirectCallback and request-login.test.ts

IceHe:
## Summary

- Simplify `parseRedirectCallback` method and `request-login.test.ts` tests
- See [discussions](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1636957110000100) for more in Slack.

## Testing

- Pass all existing tests.

### DIFF
--- a/packages/client/src/parse-callback.ts
+++ b/packages/client/src/parse-callback.ts
@@ -8,11 +8,6 @@ export interface AuthenticationResult {
 }
 
 export const parseRedirectCallback = (url: string): AuthenticationResult => {
-  const [, queryString] = url.split('?');
-  if (!queryString) {
-    return {};
-  }
-
-  const result = qs.parse(queryString);
-  return result;
+  const [, queryString = ''] = url.split('?');
+  return qs.parse(queryString);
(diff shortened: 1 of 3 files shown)


packages/client/src/parse-callback.ts
```diff
@@ -10,7 +10,7 @@ export interface AuthenticationResult {
 export const parseRedirectCallback = (url: string): AuthenticationResult => {
   const [, queryString] = url.split('?');
```
simeng-li, line 11: not sure if it is a better approach, but definitely shorter~   @wangsijie please check if this is a little bit hacky or not. 

as qs.parse('') will return empty object. 
```js
const [, queryString = ''] = url.split('?');
return qs.parse(queryString);
```

wangsijie: @IceHe woulld you mind merge this change?

IceHe: That's all right. I will do it.

wangsijie (approved):

PR #101, opened 2021-11-16
fix(client): wrap window.location.assign

wangsijie:
## Summary

fix the problem:

截屏2021-11-16 下午4 45 29

a possible reason is that returning `window.location.assign` directly will lose context

[https://stackoverflow.com/questions/10743596/why-are-certain-function-calls-termed-illegal-invocations-in-javascript](https://stackoverflow.com/questions/10743596/why-are-certain-function-calls-termed-illegal-invocations-in-javascript)


simeng-li (approved):
LGTM

PR #104, opened 2021-11-17
refactor(client): replace whitespaces in tokenSetCacheKey

IceHe:
## Summary

- Replace whitespaces in `tokenSetCacheKey` via `encodeURI` method
    - Fix related unit tests.

## Testing

Pass all existing unit tests.

### DIFF
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -24,7 +24,9 @@ const CLIENT_ID = 'client1';
 const SUBJECT = 'subject1';
 const REDIRECT_URI = 'http://localhost:3000';
 const CODE = 'code1';
-const LOGTO_TOKEN_SET_CACHE_KEY = `LOGTO_TOKEN_SET_CACHE::${ISSUER}::${CLIENT_ID}::${DEFAULT_SCOPE_STRING}`;
+const LOGTO_TOKEN_SET_CACHE_KEY = encodeURIComponent(
+  `LOGTO_TOKEN_SET_CACHE::${ISSUER}::${CLIENT_ID}::${DEFAULT_SCOPE_STRING}`
+);
 
 const discoverResponse = {
   authorization_endpoint: `${BASE_URL}/oidc/auth`,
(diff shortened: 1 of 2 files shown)


packages/client/src/index.ts
```diff
@@ -210,7 +210,9 @@ export default class LogtoClient {
   }
 
   private get tokenSetCacheKey() {
-    return `${TOKEN_SET_CACHE_KEY}::${this.oidcConfiguration.issuer}::${this.clientId}::${this.scope}`;
+    return encodeURI(
+      `${TOKEN_SET_CACHE_KEY}::${this.oidcConfiguration.issuer}::${this.clientId}::${this.scope}`
```
xiaoyijun, line 214: we just need to encode `scope` ?

packages/client/src/index.test.ts
```diff
@@ -24,7 +24,9 @@ const CLIENT_ID = 'client1';
 const SUBJECT = 'subject1';
 const REDIRECT_URI = 'http://localhost:3000';
 const CODE = 'code1';
-const LOGTO_TOKEN_SET_CACHE_KEY = `LOGTO_TOKEN_SET_CACHE::${ISSUER}::${CLIENT_ID}::${DEFAULT_SCOPE_STRING}`;
+const LOGTO_TOKEN_SET_CACHE_KEY = encodeURI(
```
wangsijie, line 27: `encodeURIComponent` would be more safe?

wangsijie (changes requested):

packages/client/src/index.ts
xiaoyijun: After disscussing with sijie, encodeURIComponent the whole key will be safer

IceHe: Fixed.

References:

- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURI
- https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/encodeURIComponent

wangsijie (approved):

PR #106, opened 2021-11-17
chore: add generateCallbackUri() method to utils

darcyYe:
## Summary
add method of generateCallbackUri() to utils

## Linear Issue Reference
[Linear LOG-238](https://linear.app/silverhand/issue/LOG-238/js-add-utils-to-generate-callback-url)

## Testing
add uni-tests to utils.test.ts and pass all the tests

### DIFF
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -7,6 +7,7 @@ import LogtoClient from '.';
 import { DEFAULT_SCOPE_STRING, SESSION_MANAGER_KEY } from './constants';
 import { MemoryStorage } from './storage';
 import { verifyIdToken } from './verify-id-token';
+import { generateCallbackUri } from './utils';
 
 const STATE = 'state1';
 
@@ -65,44 +66,6 @@ const generateIdToken = async () => {
     .sign(privateKey);
 
   return { idToken, key };
-};
-
-interface UriInfo {
-  redirectUri: string;
-  code?: string;
-  state?: string;
-  error?: string;
-  errorDescription?: string;
-}
-
-const generateCallbackUri = (uriInfo: UriInfo): string => {
-  let callbackUri = uriInfo.redirectUri.slice();
-
-  if (uriInfo.code || uriInfo.state || uriInfo.error || uriInfo.errorDescription) {
-    callbackUri += '?';
-  }
-
-  callbackUri += uriInfo.code ? `code=${uriInfo.code}` : '';
-
-  callbackUri += uriInfo.state
-    ? callbackUri.search(RegExp(/[?&]{1}$/)) != -1
-      ? `state=${uriInfo.state}`
-      : `&state=${uriInfo.state}`
-    : '';
-
-  callbackUri += uriInfo.error
-    ? callbackUri.search(RegExp(/[?&]{1}$/)) != -1
-      ? `error=${uriInfo.error}`
-      : `&error=${uriInfo.error}`
-    : '';
-
-  callbackUri += uriInfo.errorDescription
-    ? callbackUri.search(RegExp(/[?&]{1}$/)) != -1
-      ? `error_description=${uriInfo.errorDescription}`
-      : `&error_description=${uriInfo.errorDescription}`
-    : '';
-
-  return encodeURI(callbackUri);
 };
 
 describe('LogtoClient', () => {
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -6,8 +6,8 @@ import nock from 'nock';
 import LogtoClient from '.';
 import { DEFAULT_SCOPE_STRING, SESSION_MANAGER_KEY } from './constants';
 import { MemoryStorage } from './storage';
-import { verifyIdToken } from './verify-id-token';
 import { generateCallbackUri } from './utils';
+import { verifyIdToken } from './verify-id-token';
 
 const STATE = 'state1';
 
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -171,11 +171,12 @@ describe('LogtoClient', () => {
         storage,
       });
       await logto.loginWithRedirect(REDIRECT_URI, jest.fn());
-      expect(
-        logto.isLoginRedirect(
-          generateCallbackUri({ redirectUri: REDIRECT_URI, code: CODE, state: STATE })
-        )
-      ).toBeTruthy();
+      const callbackUri = generateCallbackUri({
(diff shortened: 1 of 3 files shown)


packages/client/src/index.test.ts
```diff
@@ -170,7 +171,11 @@ describe('LogtoClient', () => {
         storage,
       });
       await logto.loginWithRedirect(REDIRECT_URI, jest.fn());
-      expect(logto.isLoginRedirect(`${REDIRECT_URI}?code=${CODE}&state=${STATE}`)).toBeTruthy();
+      expect(
```
simeng-li, line 174: prefer avoid nested call stacks, just to be more readable
```js
const callbackUri = generateCallbackUri({ redirectUri: REDIRECT_URI, code: CODE, state: STATE });
expect(logto.isLoginRedirect(callbackUri)).toBeTruthy();
```

packages/client/src/utils.test.ts
```diff
@@ -32,3 +32,54 @@ describe('decodeToken', () => {
     expect(() => decodeToken(JWT)).toThrowError(StructError);
   });
 });
+
+describe('generateCallbackUri', () => {
+  test('generate callback url with redirectUri', () => {
+    const redirectUri = 'http://localhost:3000';
```
simeng-li, line 38: normally we could name using all uppercase to indicate it is a static constant. i.e.  `REDIRECT_URI`

packages/client/src/utils.test.ts
```diff
@@ -32,3 +32,54 @@ describe('decodeToken', () => {
...
+    expect(generateCallbackUri({ redirectUri })).toEqual('http://localhost:3000');
+  });
+
+  test('generate callback url with redirectUri and code', () => {
+    const redirectUri = 'http://localhost:3000';
```
simeng-li, line 43: Looks like those constants are shared cross all the test cases. We could move them to the upper scope, so no need to define multiple times.

packages/client/src/utils.ts
```diff
@@ -68,3 +76,39 @@ export const createDefaultOnRedirect = () => {
     window.location.assign(url);
   };
 };
+
+export const generateCallbackUri = (uriInfo: UriInfo): string => {
```
simeng-li, line 80: could use the destruct feature
`({ redirectUri, code, state, error, errorDescription }: UriInfo) => {} `

packages/client/src/utils.ts
```diff
@@ -68,3 +76,39 @@ export const createDefaultOnRedirect = () => {
...
+  const callbackUriBase: string = uriInfo.redirectUri
+    .slice()
+    .concat(uriInfo.code || uriInfo.state || uriInfo.error || uriInfo.errorDescription ? '?' : '');
+
+  const callbackUriConcatCode: string = callbackUriBase.concat(
+    uriInfo.code ? `code=${uriInfo.code}` : ''
+  );
+
+  const callbackUriConcatState: string = callbackUriConcatCode.concat(
```
simeng-li, line 89: we do have[ query-string](https://github.com/sindresorhus/query-string) added as a dependency. Could leverage the library to handle the search params and append it to the `redirectUri`.   

1. retrieve any original query params from the `redirectUri`
2. merge the query params with all the `non-empty` callback params
3. generate the new callbackUri

simeng-li (commented):
Please check

packages/client/src/utils.test.ts
```diff
@@ -32,3 +38,45 @@ describe('decodeToken', () => {
...
+
+describe('generateCallbackUri', () => {
+  test('generate callback url with redirectUri', () => {
+    const callbackUri = generateCallbackUri({ redirectUri: REDIRECT_URI });
+    expect(callbackUri).toEqual('http://localhost:3000');
+  });
+
+  test('generate callback url with redirectUri and code', () => {
+    const callbackUri = generateCallbackUri({ redirectUri: REDIRECT_URI, code: CODE });
+    expect(callbackUri).toEqual('http://localhost:3000?code=code1');
```
IceHe, line 50: `${REDIRECT_URI}?code=${code1}`

packages/client/src/utils.test.ts
```diff
@@ -32,3 +38,45 @@ describe('decodeToken', () => {
     expect(() => decodeToken(JWT)).toThrowError(StructError);
   });
 });
+
+describe('generateCallbackUri', () => {
+  test('generate callback url with redirectUri', () => {
+    const callbackUri = generateCallbackUri({ redirectUri: REDIRECT_URI });
+    expect(callbackUri).toEqual('http://localhost:3000');
```
IceHe, line 45: replace 'http://localhost:3000' with REDIRECT_URI

packages/client/src/utils.test.ts
```diff
@@ -32,3 +38,45 @@ describe('decodeToken', () => {
...
+
+  test('generate callback url with redirectUri, code, state and error', () => {
+    const callbackUri = generateCallbackUri({
+      redirectUri: REDIRECT_URI,
+      code: CODE,
+      state: STATE,
+      error: ERROR,
+    });
+    expect(callbackUri).toEqual(
+      'http://localhost:3000?code=code1&state=state1&error=invalid_request'
```
IceHe, line 66: ditto

packages/client/src/utils.test.ts
```diff
@@ -32,3 +38,45 @@ describe('decodeToken', () => {
...
+  test('generate callback url with full info', () => {
+    const callbackUri = generateCallbackUri({
+      redirectUri: REDIRECT_URI,
+      code: CODE,
+      state: STATE,
+      error: ERROR,
+      errorDescription: ERROR_DESCRIPTION,
+    });
+    expect(callbackUri).toEqual(
+      'http://localhost:3000?code=code1&state=state1&error=invalid_request&error_description=code_challenge%20must%20be%20a%20string%20with%20a%20minimum%20length%20of%2043%20characters'
```
IceHe, line 79: ditto

packages/client/src/utils.ts
```diff
@@ -22,6 +23,14 @@ const IDTokenSchema = s.type({
   at_hash: s.optional(s.string()),
 });
 
+interface UriInfo {
```
IceHe, line 26: TBD: Better to rename to `CallbackUriParameters`?

IceHe (changes requested):

packages/client/src/utils.ts
darcyYe: good point

IceHe (approved):
LGTM

PR #109, opened 2021-12-02
feat(react-playground): make encrypted user login info in url invisible

darcyYe:
## Summary
Encrypted user log-in info shows up in URL after logging in through react-playground. This PR fixes this problem and makes encrypted user log-in info invisible.

## Linear Issue Reference
https://linear.app/silverhand/issue/LOG-311/react-remove-query-string-in-url-after-handlecallback

## Testing
image

### DIFF
--- a/packages/react/src/provider.tsx
+++ b/packages/react/src/provider.tsx
@@ -70,7 +70,7 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
 
       dispatch({ type: 'HANDLE_CALLBACK_REQUEST' });
       try {
-        const { redirectUri = '' } = logtoClient.sessionManager.get() ?? {};
+        const { redirectUri = '' } = logtoClient.sessionManager?.get() ?? {};
         await logtoClient.handleCallback(uri);
         dispatch({ type: 'HANDLE_CALLBACK_SUCCESS', payload: { claims: logtoClient.getClaims() } });
         const [baseUrl] = redirectUri.split('?');
--- a/packages/react/src/provider.tsx
+++ b/packages/react/src/provider.tsx
@@ -73,7 +73,9 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
         const { redirectUri = '' } = logtoClient.sessionManager?.get() ?? {};
         await logtoClient.handleCallback(uri);
         dispatch({ type: 'HANDLE_CALLBACK_SUCCESS', payload: { claims: logtoClient.getClaims() } });
-        history.replaceState(history.state, '', redirectUri);
+        if (redirectUri) {
+          history.replaceState(history.state, '', redirectUri);
+        }
       } catch (error: unknown) {
         dispatch({ type: 'ERROR', payload: { error } });
       }
--- a/packages/react/src/provider.tsx
+++ b/packages/react/src/provider.tsx
@@ -1,5 +1,4 @@
 import LogtoClient, { ConfigParameters } from '@logto/client';
-import queryString from 'query-string';
 import React, { useCallback, useEffect, useReducer, useState } from 'react';
 
 import { defaultAuthState } from './auth-state';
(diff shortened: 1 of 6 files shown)


simeng-li (commented):
this should be handled in the SDK

packages/react/src/provider.tsx
```diff
@@ -72,6 +73,14 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
       try {
         await logtoClient.handleCallback(uri);
         dispatch({ type: 'HANDLE_CALLBACK_SUCCESS', payload: { claims: logtoClient.getClaims() } });
+        if (window.location.href.includes('?')) {
```
simeng-li, line 76: `if (window.location.href.includes('?'))` statement check is not necessary

packages/react/src/provider.tsx
```diff
@@ -72,6 +73,14 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
...
+          const { url: baseUrl, query: queryObject } = queryString.parseUrl(window.location.href);
```
simeng-li, line 77: Option 1: (if you want to use `query-string` as a dependency)

this whole block equals to:

```js
const filterUrl = queryString.exclude(uri, ['code', 'state', 'error', 'error_description']);
history.replaceState(history.state, '', filterUrl);
```

However we could avoid introducing extra `query-string`  dependency.
Option 2: recommended without any add-ons

```js
import { useHistory } from 'react-router-dom';

const history = useHistory();
 const queries = new URLSearchParams(window.location.search);
        
queries.delete('code');
queries.delete('state');

history.push({ search: queries.toString() });

```

simeng-li: last thing. instead of hard code the params key `code`, `state`, (`error`, `error_description` BTW), should import the keys from the `client` sdk  `AuthenticationResult` type. 

using `https://github.com/kimamula/ts-transformer-keys` or  simply define a const array but type is guarded by keyof `AuthenticationResult`

simeng-li (changes requested):

packages/react/src/provider.tsx
```diff
@@ -72,6 +73,14 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
...
+          if (baseUrl && Object.keys(queryObject).length > 0) {
+            const { code, state, ...filteredEntries } = queryObject;
+            const newUrl = queryString.stringifyUrl({ url: baseUrl, query: filteredEntries });
```
wangsijie, line 80: The if statement and querystring parse & stringify are unnacessary:
1. we know that current `location.href` a valid callback url with code and state, or we won't call this function (`handleCallback`)
2. the `redirectUri` is already set and we know exactly what it is, no need to recover query entries.
cc @simeng-li

wangsijie (changes requested):

packages/react/src/provider.tsx
simeng-li: make sense

packages/react/src/provider.tsx
```diff
@@ -73,14 +72,8 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
...
-        if (window.location.href.includes('?')) {
-          const { url: baseUrl, query: queryObject } = queryString.parseUrl(window.location.href);
-          if (baseUrl && Object.keys(queryObject).length > 0) {
-            const { code, state, ...filteredEntries } = queryObject;
-            const newUrl = queryString.stringifyUrl({ url: baseUrl, query: filteredEntries });
-            history.replaceState(history.state, '', newUrl);
-          }
-        }
+        const [baseUrl] = uri.split('?');
+        history.replaceState(history.state, '', baseUrl);
```
wangsijie, line 76: `redirectUri` can be found in client's session manager, but should be fetched before `handleCallback`.

wangsijie (changes requested):

packages/react/src/provider.tsx
```diff
@@ -70,8 +70,11 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
 
       dispatch({ type: 'HANDLE_CALLBACK_REQUEST' });
       try {
+        const { redirectUri = '' } = logtoClient.sessionManager?.get() ?? {};
         await logtoClient.handleCallback(uri);
         dispatch({ type: 'HANDLE_CALLBACK_SUCCESS', payload: { claims: logtoClient.getClaims() } });
+        const [baseUrl] = redirectUri.split('?');
```
simeng-li, line 76: do we need to remove the original query params?

packages/react/src/provider.tsx
```diff
@@ -70,8 +70,10 @@ export const LogtoProvider = ({ logtoConfig, children }: LogtoProviderProperties
 
       dispatch({ type: 'HANDLE_CALLBACK_REQUEST' });
       try {
+        const { redirectUri = '' } = logtoClient.sessionManager?.get() ?? {};
         await logtoClient.handleCallback(uri);
         dispatch({ type: 'HANDLE_CALLBACK_SUCCESS', payload: { claims: logtoClient.getClaims() } });
+        history.replaceState(history.state, '', redirectUri);
```
wangsijie, line 76: nit: if `redirectUri` is empty, stop replaceState, although it won't happen theoretically.

wangsijie (approved):

packages/react/src/provider.tsx
darcyYe: make sense

simeng-li (approved):

PR #110, opened 2021-12-07
refactor(client): remove onAuthStateChange callback method

simeng-li:
## Summary

remove onAuthStateChange callback method of js SDK as it is no longer needed

## Testing

UT case passed and related test case removed. 

@wangsijie


darcyYe (approved):
LGTM

PR #117, opened 2022-01-06
refactor(client): decodeIdToken

IceHe:
## Summary

- Rename `IDToken` to `IdTokenClaims`
    - Reference: [IdTokenClaims](https://www.notion.so/silverhand/9c1fbcd630fc44adaa45a6d7d686326a?v=3cea92e2906048d287a8648b76c73c65)
- Rename `decodeToken` to `decodeIdToken`
    - Reference: [decodeIdToken](https://www.notion.so/silverhand/decodeIdToken-7b3b739c397a431882a2e41bac8782f1)
- Add necessary optional claims to `IdTokenClaims`
    - Reference: [ID Token Claims](https://www.notion.so/silverhand/OIDC-ec922f42788a41a69f5e2550a2f7c3fe#4ffdd579db9b473d8876ce870fcdaede)
- Update code related to ID Token

Notes:

- The standard claims will be added later when implementing `UserInfoResponse` of `fetchUserInfo`.
    - Reference: [Standard Claims](https://openid.net/specs/openid-connect-core-1_0.html#StandardClaims)

## Testing

Pass all existing tests.

---

@logto-io/eng

### DIFF
--- a/packages/client/src/modules/token-set.ts
+++ b/packages/client/src/modules/token-set.ts
@@ -1,6 +1,6 @@
 import { TokenResponse } from '../api';
 import { nowRoundToSec } from '../utils';
-import { decodeToken, IDToken } from '../utils/id-token';
+import { decodeIdToken, IdTokenClaims } from '../utils/id-token';
 
 export default class TokenSet {
   public accessToken: string;
@@ -26,11 +26,11 @@ export default class TokenSet {
     return this.expiresIn === 0;
   }
 
-  public claims(): IDToken {
+  public claims(): IdTokenClaims {
     if (!this.idToken) {
       throw new TypeError('id_token not present in TokenSet');
     }
 
-    return decodeToken(this.idToken);
+    return decodeIdToken(this.idToken);
   }
 }
--- a/packages/client/src/utils/id-token.test.ts
+++ b/packages/client/src/utils/id-token.test.ts
@@ -4,10 +4,10 @@ import { SignJWT, generateKeyPair } from 'jose';
 import nock from 'nock';
 import { StructError } from 'superstruct';
 
-import { decodeToken, createJWKS, verifyIdToken } from './id-token';
+import { decodeIdToken, createJWKS, verifyIdToken } from './id-token';
 
 describe('verifyIdToken', () => {
-  test('valid idToken', async () => {
+  test('valid ID Token', async () => {
     const { privateKey, publicKey } = await generateKeyPair('RS256');
 
     if (!(publicKey instanceof KeyObject)) {
@@ -107,8 +107,8 @@ describe('verifyIdToken', () => {
   });
 });
 
-describe('decodeToken', () => {
-  test('decode token and get claims', async () => {
+describe('decodeIdToken', () => {
+  test('decode ID Token and get claims', async () => {
     const { privateKey } = await generateKeyPair('RS256');
     const JWT = await new SignJWT({})
       .setProtectedHeader({ alg: 'RS256' })
@@ -118,15 +118,15 @@ describe('decodeToken', () => {
       .setIssuedAt()
       .setExpirationTime('2h')
       .sign(privateKey);
-    const payload = decodeToken(JWT);
-    expect(payload.sub).toEqual('foz');
+    const idTokenClaims = decodeIdToken(JWT);
+    expect(idTokenClaims.sub).toEqual('foz');
   });
 
   test('throw on invalid JWT string', async () => {
-    expect(() => decodeToken('invalid-JWT')).toThrow();
+    expect(() => decodeIdToken('invalid-JWT')).toThrow('invalid token');
   });
 
-  test('throw ZodError when iss is missing', async () => {
+  test('throw StructError when iss is missing', async () => {
     const { privateKey } = await generateKeyPair('RS256');
     const JWT = await new SignJWT({})
       .setProtectedHeader({ alg: 'RS256' })
@@ -135,6 +135,6 @@ describe('decodeToken', () => {
       .setIssuedAt()
       .setExpirationTime('2h')
       .sign(privateKey);
-    expect(() => decodeToken(JWT)).toThrowError(StructError);
+    expect(() => decodeIdToken(JWT)).toThrowError(StructError);
   });
 });
--- a/packages/client/src/utils/id-token.ts
+++ b/packages/client/src/utils/id-token.ts
@@ -15,25 +15,31 @@ const fullfillBase64 = (input: string) => {
   return input;
 };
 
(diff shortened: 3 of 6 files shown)


packages/client/src/modules/token-set.ts
```diff
@@ -1,6 +1,6 @@
 import { TokenResponse } from '../api';
 import { nowRoundToSec } from '../utils';
-import { decodeToken, IDToken } from '../utils/id-token';
+import { decodeIdToken, IdTokenClaims } from '../utils/id-token';
```
gao-sun, line 3: should add `@` alias?

packages/client/src/utils/id-token.test.ts
```diff
@@ -118,12 +118,12 @@ describe('decodeToken', () => {
...
+    const idTokenClaims = decodeIdToken(JWT);
+    expect(idTokenClaims.sub).toEqual('foz');
   });
 
   test('throw on invalid JWT string', async () => {
-    expect(() => decodeToken('invalid-JWT')).toThrow();
+    expect(() => decodeIdToken('invalid-JWT')).toThrow();
   });
 
   test('throw ZodError when iss is missing', async () => {
```
gao-sun, line 129: should be `StructError`?

packages/client/src/modules/token-set.ts
```diff
@@ -26,11 +26,11 @@ export default class TokenSet {
     return this.expiresIn === 0;
   }
 
-  public claims(): IDToken {
+  public claims(): IdTokenClaims {
```
gao-sun, line 29: will it be better to use [property getter](https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Functions/get)?

packages/client/src/utils/id-token.ts
```diff
@@ -15,25 +15,31 @@ const fullfillBase64 = (input: string) => {
...
 };
 
-const IDTokenSchema = s.type({
+const IdTokenClaimsSchema = s.type({
   iss: s.string(),
   sub: s.string(),
   aud: s.string(),
   exp: s.number(),
   iat: s.number(),
+  auth_time: s.optional(s.number()),
```
simeng-li, line 24: do we need all these fields declared?

packages/client/src/modules/token-set.ts
simeng-li: I believe there should be a separate PR related to the tokenSet refactor. cc: @IceHe

packages/client/src/utils/id-token.ts
IceHe: I find that some of these claims seems to be unused for now,
so I can also do not add them.

But I still added them here according to the documentations [decodeIdToken](https://www.notion.so/silverhand/decodeIdToken-7b3b739c397a431882a2e41bac8782f1) and [ID Token](https://www.notion.so/silverhand/OIDC-ec922f42788a41a69f5e2550a2f7c3fe#4ffdd579db9b473d8876ce870fcdaede)

@wangsijie @xiaoyijun any ideas? Add them now or when needed?

IceHe:
@simeng-li ok, I will refactor TokenSet in another PR according to [documentation](https://www.notion.so/silverhand/decodeIdToken-7b3b739c397a431882a2e41bac8782f1):

- Make decodeIdToken just return a object in `js/js` (core)
- Convert this object to IdTokenClaims type in the platform SDKs.

packages/client/src/utils/id-token.test.ts
IceHe: `decodeIdToken` throws `StructError` (imported via SuperStruct) only when SuperStruct `assert` throw `StructError`, otherwise it throws a normal `Error` for now.

```ts
export const decodeIdToken = (token: string): IDToken => {
  const payloadPart = token.split('.')[1];

  if (!payloadPart) {
    throw new Error('invalid token');
  }

  const payloadString = payloadPart.replace(/-/g, '+').replace(/_/g, '/');
  const json = decodeURIComponent(
    escape(Buffer.from(fullfillBase64(payloadString), 'base64').toString())
  );

  try {
    // Using SuperStruct to validate the json type
    const idToken = JSON.parse(json) as IDToken;
    s.assert(idToken, IDTokenSchema);
    return idToken;
  } catch (error: unknown) {
    if (error instanceof s.StructError) {
      throw error;
    }

    throw new Error('invalid token: JSON parse failed');
  }
};
```

Should we throw `StructError` instead of normal `Error` all the time here?

If any string cannot be decoded to ID Token, we seems reasonable to assert that there is a struct error in this string.

packages/client/src/modules/token-set.ts
IceHe: Ok, I will do it later in another PR. [issue](https://linear.app/silverhand/issue/LOG-1275/use-alias-path-in-js-project)

packages/client/src/utils/id-token.test.ts
IceHe: Here is a normal `Error`.
I will assert its content later

IceHe: You mean `decodeIdToken` should not throw `StructError`?

packages/client/src/utils/id-token.ts
IceHe: I think we'd better declare them, which will be used in the near future.

gao-sun: maybe optional?

packages/client/src/modules/token-set.ts
gao-sun: @IceHe is it true? please paste linear issue number here before resolving the conversation

packages/client/src/modules/token-set.ts
gao-sun: you can directly use LOG-1275 to refer since we have auto-link configured

packages/client/src/modules/token-set.ts
IceHe: - https://linear.app/silverhand/issue/LOG-1272/browser-refactor
    - https://linear.app/silverhand/issue/LOG-1273/refactor-tokenset

packages/client/src/modules/token-set.ts
IceHe: Nice feature. Got it.

packages/client/src/modules/token-set.ts
gao-sun: mind to elaborate LOG-1273 in case we forget the context? you can specify what to refactor or paste PR comment link

IceHe: Done

simeng-li (approved):
LGTM

PR #119, opened 2022-01-12
refactor: unit tests of logto client

IceHe:
## Summary

Refactor LogtoClient unit tests

- Mock core functions and utility functions.
- Simplify generateIdToken
- Remove the unnecessary mocks

Background:

- ~~Some LogtoClient unit tests are actually used to test core and util functions. As these unit tests make it hard to split `js/client` into `js/js` and `js/browser`, I have to refactor them at first.~~
- But now I know it's just because I didn't know how to easily mock the function from another package at that time.

Next up:

- Related updates, including renaming variables, modifying description of test cases and formatting code, will be committed in my next PR.

## Linear Issue Reference

LOG-1296

## Testing

Pass all existing tests.

### Coverage

image

Branches coverage of `src/api` will be improved (within the linear issues about core functions) in the future.

---
@logto-io/eng

### DIFF
--- a/packages/client/src/index.test.ts
+++ b/packages/client/src/index.test.ts
@@ -1,26 +1,15 @@
 import { KeyObject } from 'crypto';
 
 import { generateKeyPair, SignJWT } from 'jose';
-import nock from 'nock';
 
 import LogtoClient from '.';
+import { discover, grantTokenByAuthorizationCode, grantTokenByRefreshToken } from './api';
 import { DEFAULT_SCOPE_STRING, SESSION_MANAGER_KEY } from './constants';
 import { MemoryStorage } from './modules/storage';
-import { verifyIdToken } from './utils/id-token';
+import { getLoginUrlWithCodeVerifierAndState, getLogoutUrl } from './utils/assembler';
+import { createJWKS, verifyIdToken } from './utils/id-token';
+import { createRequester } from './utils/requester';
 import { generateCallbackUri } from './utils/utils-test';
-
-const STATE = 'state1';
-
-jest.mock('./utils/generators', () => ({
-  ...jest.requireActual('./utils/generators'),
-  generateState: () => STATE,
-}));
-
-jest.mock('./utils/id-token', () => ({
-  ...jest.requireActual('./utils/id-token'),
-  verifyIdToken: jest.fn(),
-  createJWKS: jest.fn(),
-}));
 
 const DOMAIN = 'logto.dev';
 const BASE_URL = `https://${DOMAIN}`;
@@ -29,6 +18,9 @@ const CLIENT_ID = 'client1';
 const SUBJECT = 'subject1';
 const REDIRECT_URI = 'http://localhost:3000';
 const CODE = 'code1';
+const CODE_VERIFIER = 'codeVerifier1';
+const STATE = 'state1';
+
 const LOGTO_TOKEN_SET_CACHE_KEY = encodeURIComponent(
   `LOGTO_TOKEN_SET_CACHE::${ISSUER}::${CLIENT_ID}::${DEFAULT_SCOPE_STRING}`
 );
@@ -42,13 +34,6 @@ const discoverResponse = {
   end_session_endpoint: `${BASE_URL}/oidc/session/end`,
 };
 
-const fakeTokenResponse = {
-  access_token: 'access_token',
-  expires_in: 3600,
-  id_token: 'id_token',
-  refresh_token: 'refresh_token',
-};
-
 const generateIdToken = async () => {
   const { privateKey, publicKey } = await generateKeyPair('RS256');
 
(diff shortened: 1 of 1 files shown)


packages/client/src/index.test.ts
```diff
@@ -1,470 +1,519 @@
...
+    issuer: ISSUER,
+    authorization_endpoint: `${ISSUER}/authorize`,
+    jwks_uri: `${ISSUER}/jwks`,
+    token_endpoint: `${ISSUER}/token`,
+    revocation_endpoint: `${ISSUER}/revoke_token`,
+    end_session_endpoint: `${ISSUER}/end_session`,
+  }));
+
+  const grantTokenByAuthorizationCode = jest
+    .fn()
```
wangsijie, line 65:
```ts
fn(async () => Promise.resolve(TOKEN_RESPONSE));
```

packages/client/src/index.test.ts
```diff
@@ -1,470 +1,519 @@
...
+  const getSignOutUrl = jest.fn().mockReturnValue(`${ISSUER}/sign_out`);
+
+  return {
+    getLoginUrlWithCodeVerifierAndState: getSignInUrlWithCodeVerifierAndState,
+    getLogoutUrl: getSignOutUrl,
+  };
+});
 
-  describe('createLogtoClient', () => {
+jest.mock('./utils/id-token', () => ({
```
wangsijie, line 94: the second arg is useless

packages/client/src/index.test.ts
```diff
@@ -1,470 +1,519 @@
...
+    });
+
+    test('discover and createRequester should have been called', async () => {
+      await LogtoClient.create({
+        domain: DOMAIN,
+        clientId: CLIENT_ID,
+      });
+
+      expect(discover).toBeCalled();
+      expect(createRequester).toBeCalled();
```
wangsijie, line 121: do we need to align with `toHaveBeenCalled`?

wangsijie (changes requested):

wangsijie (approved):

simeng-li:
nit: for future references, we'd better separate naming updates with logical ones. Sometimes it's a little bit hard to identify the extract purpose of some individual updates without some context.

IceHe:
Other updates, including renaming variables, modifying description of test cases and formatting code, will be committed in my next PR.

simeng-li (commented):
please check. Also, we'd better keep test file updates along with product file updates within a single PR, so we may have a better context of how the original test cases break and why we need to update the test case. By default, a test case shouldn't be updated till a feature update that may break current ones.

packages/client/src/index.test.ts
```diff
@@ -42,49 +31,68 @@ const discoverResponse = {
...
+    discover,
+    grantTokenByAuthorizationCode,
+    grantTokenByRefreshToken,
+  };
+});
+
+const CODE_VERIFIER = 'codeVerifier1';
+const STATE = 'state1';
 
+jest.mock('./utils/assembler', () => {
```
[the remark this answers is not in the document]
IceHe: Need to mock because:

-   We just require it to return specific value as we expected here
 
    If we call the actual `getLoginUrlWithCodeVerifierAndState`,
    we have to mock `generateState` and `generateCodeVerifier` instead
    so that we force `getLoginUrlWithCodeVerifierAndState` to return same url, codeVerifier and state.
    
-   `getLoginUrlWithCodeVerifierAndState` should be completely tested in `assembler.test.ts` and then we just care what it does but not how to implement it.

    I'd better mock it here.

For reference
  
```ts
export const getLoginUrlWithCodeVerifierAndState = async (
  parameters: LoginPrepareParameters
): Promise<{ url: string; codeVerifier: string; state: string }> => {
  const { baseUrl, clientId, scope, redirectUri } = parameters;
  const codeVerifier = generateCodeVerifier();
  const codeChallenge = await generateCodeChallenge(codeVerifier);
  const state = generateState();

  const url = `${baseUrl}?${qs.stringify({
    client_id: clientId,
    scope,
    response_type: 'code',
    redirect_uri: [redirectUri],
    prompt: 'consent',
    code_challenge: codeChallenge,
    code_challenge_method: 'S256',
    state,
  })}`;

  return { url, codeVerifier, state };
};
```

IceHe:
> please check. Also, we'd better keep test file updates along with product file updates within a single PR, so we may have a better context of how the original test cases break and why we need to update the test case. By default, a test case shouldn't be updated till a feature update that may break current ones.

U'r right.

I used to think that it's hard to split because LogtoClient unit tests are not reasonable enough.
But I realized yesterday that it's just because I didn't know how to easily mock the function from another package at that time.

( LogtoClient unit tests are not reasonable but they're exactly unnecessary to refactor for now. )

Merge or close (do it in the future)? (I prefer to merge)

gao-sun:
is there a chance we can ignore those tests and do it from scratch? looks like the ROI is too low if we consider the old tests when refactoring.

gao-sun:
also the single test file is ~500 LOC, which is not acceptable. @simeng-li @wangsijie please take a look at this issue, also make sure we have related lint rule to limit 300 LOC for `.ts` and 350 LOC for `.tsx`.

IceHe:
Do it in another way: split js/client via moving the core and util functions from js/client to js/js one by one and refactor the LogtoClient unit tests at the same time

PR #121, opened 2022-01-13
refactor!: split js/client into js/js and js/browser

IceHe:
## Summary

- refactor!: split js/client into js/js and js/browser

## Linear Issue Reference

LOG-1266

## Testing

Pass all existing tests


IceHe:
Ditto #119

PR #125, opened 2022-01-18
feat(js): generate state, code verifier and code challenge

IceHe:
## Summary

- Add generators
    - generateState
    - generateCodeVerifier
    - generateCodeChallenge

## Linear Issue Reference

LOG-1354

## Testing

- generateState:
    - result should be random value
- generateCodeVerifier
    - result should be random value
- generateCodeChallenge
    - should generate different codeChallenge with different code verifiers
    - should generate same codeChallenge with same code verifiers

---
@logto-io/eng

### DIFF
--- a/packages/js/src/utils/generators.test.ts
+++ b/packages/js/src/utils/generators.test.ts
@@ -1,0 +1,84 @@
+/**
+ * @jest-environment node
+ */
+import { toUint8Array } from 'js-base64';
+
+import { generateCodeChallenge, generateCodeVerifier, generateState } from './generators';
+
+const isUrlSafe = (input: string) => /^[\w-]*$/.test(input);
+
+describe('generateState', () => {
+  test('should be random value', () => {
+    const state1 = generateState();
+    const state2 = generateState();
+    expect(state1).not.toEqual(state2);
+  });
+
+  test('should be url-safe', () => {
+    const state = generateState();
+    expect(isUrlSafe(state)).toBeTruthy();
+  });
+
+  test('raw random data length should be length 64', () => {
+    const state = generateState();
+    expect(toUint8Array(state).length).toEqual(64);
+  });
+});
+
+describe('generateCodeVerifier', () => {
+  test('should be random value', () => {
+    const codeVerifier1 = generateCodeVerifier();
+    const codeVerifier2 = generateCodeVerifier();
+    expect(codeVerifier1).not.toEqual(codeVerifier2);
+  });
+
+  test('should be url-safe', () => {
+    const codeVerifier = generateCodeVerifier();
+    expect(isUrlSafe(codeVerifier)).toBeTruthy();
+  });
+
+  test('raw random data length should be length 64', () => {
+    const codeVerifier = generateCodeVerifier();
+    expect(toUint8Array(codeVerifier).length).toEqual(64);
+  });
+});
+
+describe('generateCodeChallenge', () => {
+  test('dealing with different code verifiers should not be equal', async () => {
+    const codeVerifier1 = generateCodeVerifier();
+    const codeChallenge1 = await generateCodeChallenge(codeVerifier1);
+    const codeVerifier2 = generateCodeVerifier();
+    const codeChallenge2 = await generateCodeChallenge(codeVerifier2);
+    expect(codeChallenge1).not.toEqual(codeChallenge2);
+  });
+
+  test('dealing with same code verifier should be equal', async () => {
+    const codeVerifier = generateCodeVerifier();
+    const codeChallenge1 = await generateCodeChallenge(codeVerifier);
+    const codeChallenge2 = await generateCodeChallenge(codeVerifier);
+    expect(codeChallenge1).toEqual(codeChallenge2);
+  });
+
+  describe('dealing with static code verifier should not throw', () => {
+    test('dealing with url-safe code verifier should not throw', async () => {
+      const codeVerifier =
+        'tO6MabnMFRAatnlMa1DdSstypzzkgalL1-k8Hr_GdfTj-VXGiEACqAkSkDhFuAuD8FOU8lMishaXjt29Xt2Oww';
+      const codeChallenge = await generateCodeChallenge(codeVerifier);
+      expect(codeChallenge).toEqual('0K3SLeGlNNzFswYJjcVzcN4C76m_8NZORxFJLBJWGwg');
+    });
+
+    describe('dealing with non-url-safe code verifier should not throw', () => {
+      test('latin1 character', async () => {
+        const codeVerifier = 'Á';
+        const codeChallenge = await generateCodeChallenge(codeVerifier);
+        expect(codeChallenge).toEqual('p3yvZiKYauPicLIDZ0W1peDz4Z9KFC-9uxtDfoO1KOQ');
+      });
+
+      test('emoji character', async () => {
+        const codeVerifier = '🚀';
+        const codeChallenge = await generateCodeChallenge(codeVerifier);
+        expect(codeChallenge).toEqual('67wLKHDrMj8rbP-lxJPO74GufrNq_HPU4DZzAWMdrsU');
+      });
+    });
+  });
+});
--- a/packages/js/src/utils/generators.ts
+++ b/packages/js/src/utils/generators.ts
@@ -1,0 +1,33 @@
+/** @link [Proof Key for Code Exchange by OAuth Public Clients](https://datatracker.ietf.org/doc/html/rfc7636) */
+
+import { fromUint8Array } from 'js-base64';
+
+/**
+ * @param length The length of the raw random data.
+ */
+const generateRandomString = (length = 64) =>
+  fromUint8Array(crypto.getRandomValues(new Uint8Array(length)), true);
+
+/**
+ * Generates random string for state and encodes them in url safe base64
+ */
+export const generateState = () => generateRandomString();
+
+/**
+ * Generates code verifier
+ *
+ * @link [Client Creates a Code Verifier](https://datatracker.ietf.org/doc/html/rfc7636#section-4.1)
+ */
+export const generateCodeVerifier = () => generateRandomString();
+
+/**
+ * Calculates the S256 PKCE code challenge for an arbitrary code verifier and encodes it in url safe base64
+ *
+ * @param {String} codeVerifier Code verifier to calculate the S256 code challenge for
+ * @link [Client Creates the Code Challenge](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)
+ */
+export const generateCodeChallenge = async (codeVerifier: string): Promise<string> => {
+  const encodedCodeVerifier = new TextEncoder().encode(codeVerifier);
+  const codeChallenge = new Uint8Array(await crypto.subtle.digest('SHA-256', encodedCodeVerifier));
+  return fromUint8Array(codeChallenge, true);
+};
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -18,7 +18,8 @@
     "report": "WITH_REPORT=true npm run package"
   },
   "dependencies": {
-    "@silverhand/essentials": "^1.1.2"
+    "@silverhand/essentials": "^1.1.2",
+    "js-base64": "^3.7.2"
   },
   "devDependencies": {
     "@peculiar/webcrypto": "^1.1.7",
--- a/packages/js/src/index.test.ts
+++ b/packages/js/src/index.test.ts
@@ -1,9 +1,0 @@
-// Temporary content, for `pnpm test` to pass CI, will be deleted later
-import { TODO } from '.';
-
-describe('TODO', () => {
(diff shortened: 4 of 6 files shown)


packages/js/src/utils/generators.test.ts
```diff
@@ -0,0 +1,69 @@
...
+    const verifier2 = generateCodeVerifier();
+    expect(verifier1).not.toEqual(verifier2);
+  });
+});
+
+describe('generateCodeChallenge', () => {
+  test('with fixed length', async () => {
+    const verifier = generateCodeVerifier();
+    const challenge = await generateCodeChallenge(verifier);
+    expect(challenge.length).toEqual(Math.ceil(32 * 1.34));
```
simeng-li, line 52: nit: could have a comment for these magic numbers.

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,58 @@
...
+
+/**
+ * Generates random bytes for state and encodes them in url safe base64
+ */
+export const generateState = () => generateRandomString(CODE_VERIFIER_MAX_LENGTH);
+
+/**
+ * Generates random bytes for nonce and encodes them in url safe base64
+ */
+export const generateNonce = () => generateRandomString(CODE_VERIFIER_MAX_LENGTH);
```
simeng-li, line 39: Do we cover this method?  Looks like missing from the conversion doc

simeng-li (commented):
please check

packages/js/src/utils/generators.ts
IceHe: I'm good with pre-defining unused utility function for now.
(add it according to the implementation of `oidc-client`)

@xiaoyijun any ideas?

- A. remove for now
- B. reserve it

xiaoyijun: A

packages/js/src/utils/generators.test.ts
IceHe: Added, PTAL.

IceHe:
Removed the dependency of `nanoid`

packages/js/src/utils/generators.test.ts
gao-sun: this line is not expected. we do not care the length of the generated string

packages/js/src/utils/generators.test.ts
```diff
@@ -0,0 +1,58 @@
...
+  generateCodeChallenge,
+  generateCodeVerifier,
+  generateState,
+} from './generators';
+
+describe('generateState', () => {
+  test('with fixed length', () => {
+    const state = generateState();
+    expect(state.length).toEqual(CODE_VERIFIER_MAX_LENGTH);
+  });
```
gao-sun, line 12: all these `with fixed length` s are not expected

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,53 @@
...
+import { UrlSafeBase64 } from '@silverhand/essentials';
+import { fromUint8Array } from 'js-base64';
+
+/**
+ * Length Range of Code Verifier
+ *
+ * The length of code verifier ranges from 43 to 128.
+ * @link [Client Creates a Code Verifier](https://datatracker.ietf.org/doc/html/rfc7636#section-4.1)
+ */
+export const CODE_VERIFIER_MAX_LENGTH = 128;
```
gao-sun, line 12: why do we need this constant? our sdk convention specified we hard code all random string in a raw length of 64.

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,53 @@
...
+
+/**
+ * The strings of length 96 will contain 128 characters after encoding in base64.
+ * Because the base64 encoding cause an overhead of 33-36%.
+ * @link [Base64](Reference: https://en.wikipedia.org/wiki/Base64)
+ */
+const CODE_VERIFIER_RAW_RANDOM_DATA_LENGTH = 96;
```
gao-sun, line 19: this doesn't fit our sdk convention..

packages/js/src/utils/generators.test.ts
```diff
@@ -0,0 +1,58 @@
...
+
+  test('with random value', () => {
+    const verifier1 = generateCodeVerifier();
+    const verifier2 = generateCodeVerifier();
+    expect(verifier1).not.toEqual(verifier2);
+  });
+});
+
+describe('generateCodeChallenge', () => {
```
gao-sun, line 34: you also need some static [test cases](https://github.com/logto-io/swift/blob/master/Tests/LogtoTests/LogtoUtilitiesTests.swift#L18-L38) to prove the algorithm is correct, actually that's the core

gao-sun (changes requested):

packages/js/src/utils/generators.test.ts
IceHe: Removed.

packages/js/src/utils/generators.ts
IceHe: Removed.

packages/js/src/utils/generators.ts
IceHe: Removed.

packages/js/src/utils/generators.test.ts
IceHe: Removed.

packages/js/src/utils/generators.test.ts
IceHe: Added.

packages/js/src/utils/generators.test.ts
```diff
@@ -0,0 +1,77 @@
...
+
+  test('dealing with same code verifier should be equal', async () => {
+    const codeVerifier =
+      'KH5wVipNo1W0cpKnx0csh_WlmY5fpmAiYl46OBBc2SE6j93RtKo8j7M230jQ9DF0IWBFl1j_UsrI_5dcpHdo1w';
+    const codeChallenge1 = await generateCodeChallenge(codeVerifier);
+    const codeChallenge2 = await generateCodeChallenge(codeVerifier);
+    expect(codeChallenge1).toEqual(codeChallenge2);
+  });
+
+  test('dealing with code verifier containing latin1 character', async () => {
```
IceHe, line 66: Discuss: is it necessary test cases?
(added according to [unit tests](https://github.com/logto-io/swift/blob/master/Tests/LogtoTests/LogtoUtilitiesTests.swift#L18-L38) from Swift SDK)

IceHe: @logto-io/eng any idea?

packages/js/src/utils/generators.test.ts
```diff
@@ -0,0 +1,77 @@
...
+    const codeVerifier = 'Á';
+    const codeChallenge = await generateCodeChallenge(codeVerifier);
+    expect(codeChallenge).toEqual('p3yvZiKYauPicLIDZ0W1peDz4Z9KFC-9uxtDfoO1KOQ');
+  });
+
+  test('dealing with code verifier containing emoji character', async () => {
```
IceHe, line 72: ditto

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,39 @@
...
+import { fromUint8Array } from 'js-base64';
+
+const DEFAULT_RANDOM_DATA_LENGTH = 64;
+
+/**
+ * @param length The length of the raw random data.
+ */
+function generateRandomString(length: number) {
+  const randomValues = crypto.getRandomValues(new Uint8Array(length));
+  const base64String = fromUint8Array(randomValues);
```
IceHe, line 13: If use `crypto.getRandomValues(new Uint8Array(length))`, which returns `Uint8Array` data, to generate random data, how to implement base64-encode and base64-decode under both Node and browser-like environments?

( js/js should be able to run under both Node and browser-like environments)

- from `Uint8Array` to base64 string
- from base64 string to `Uint8Array`

Any idea? Ask for help @logto-io/eng

IceHe: image

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,39 @@
...
+/**
+ * Calculates the S256 PKCE code challenge for an arbitrary code verifier and encodes it in url safe base64
+ *
+ * @param {String} codeVerifier Code verifier to calculate the S256 code challenge for
+ * @link [Client Creates the Code Challenge](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)
+ */
+export const generateCodeChallenge = async (codeVerifier: string): Promise<string> => {
+  const encoded = new TextEncoder().encode(codeVerifier);
+  const challenge = new Uint8Array(await crypto.subtle.digest('SHA-256', encoded));
+  return UrlSafeBase64.replaceNonUrlSafeCharacters(fromUint8Array(challenge));
```
simeng-li, line 38: Looks like `fromUint8Array` takes a second param which returns a URI safe  Base64

IceHe: Fixed, PTAL.

IceHe: Again.

packages/js/src/utils/generators.test.ts
```diff
@@ -0,0 +1,76 @@
+import { toUint8Array } from 'js-base64';
+
+import { generateCodeChallenge, generateCodeVerifier, generateState } from './generators';
+
+const isUrlSafe = (input: string) => /^[\w-]*$/.test(input);
```
gao-sun, line 5: we can add this into `essentials`

packages/js/src/utils/generators.test.ts
gao-sun: i didn't mean 'dealing with different code verifiers should not be equal' and 'dealing with same code verifier should be equal' should use static strings.. 😂 should add a new `test()` to test static values, please see the link above

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,36 @@
+/** @link [Proof Key for Code Exchange by OAuth Public Clients](https://datatracker.ietf.org/doc/html/rfc7636) */
+
+import { fromUint8Array } from 'js-base64';
+
+const DEFAULT_RANDOM_DATA_LENGTH = 64;
+
+/**
+ * @param length The length of the raw random data.
+ */
+function generateRandomString(length: number) {
```
gao-sun, line 10:
```suggestion
const generateRandomString = (length = DEFAULT_RANDOM_DATA_LENGTH) =>
```

gao-sun: also we do not have the convention to use UPPER_CASE for constants in js. camelCase would be fine.

packages/js/src/utils/generators.ts
```diff
@@ -0,0 +1,36 @@
...
+/**
+ * Calculates the S256 PKCE code challenge for an arbitrary code verifier and encodes it in url safe base64
+ *
+ * @param {String} codeVerifier Code verifier to calculate the S256 code challenge for
+ * @link [Client Creates the Code Challenge](https://datatracker.ietf.org/doc/html/rfc7636#section-4.2)
+ */
+export const generateCodeChallenge = async (codeVerifier: string): Promise<string> => {
+  const encodedCodeVerifier = new TextEncoder().encode(codeVerifier);
+  const codeChallenge = new Uint8Array(await crypto.subtle.digest('SHA-256', encodedCodeVerifier));
+  return fromUint8Array(codeChallenge, true);
```
gao-sun, line 35: we can remove related func in `essentials` if needed

packages/js/src/utils/generators.test.ts
gao-sun: no hurries, you can add a linear issue and do it later

packages/js/src/utils/generators.ts
gao-sun: no hurries too

packages/js/src/utils/generators.test.ts
IceHe: use `UrlSafeBase64` in #128 , you can take a look

packages/js/src/utils/generators.ts
IceHe: ditto: use UrlSafeBase64 in #128 , you can take a look

packages/js/src/utils/generators.ts
IceHe: Fixed, PTAL.

packages/js/src/utils/generators.test.ts
IceHe: Fixed, PTAL.

packages/js/src/utils/generators.test.ts
gao-sun: ok

gao-sun (approved):
lgtm

packages/js/src/utils/generators.test.ts
IceHe: Added issue LOG-1378 and will do it later

PR #126, opened 2022-01-18
feat(js): decodeIdToken

IceHe:
## Summary

- Add utility function `decodeIdToken`
- Add basic type `IdTokenClaims`

Related #117 

## Linear Issue Reference

LOG-1355
LOG-1357

## Testing

- decode valid JWT → return claims
- decode invalid JWT → throw Error
- decode valid JWT without `iss` → throw StructError

---
@logto-io/eng

### DIFF
--- a/packages/js/src/utils/id-token.test.ts
+++ b/packages/js/src/utils/id-token.test.ts
@@ -1,0 +1,42 @@
+import { SignJWT, generateKeyPair } from 'jose';
+import { StructError } from 'superstruct';
+
+import { decodeIdToken } from './id-token';
+
+describe('decodeIdToken', () => {
+  test('decoding valid JWT should return claims', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const jwt = await new SignJWT({})
+      .setProtectedHeader({ alg: 'RS256' })
+      .setIssuer('foo')
+      .setSubject('bar')
+      .setAudience('qux')
+      .setExpirationTime(2000)
+      .setIssuedAt(1000)
+      .sign(privateKey);
+    const idTokenClaims = decodeIdToken(jwt);
+    expect(idTokenClaims).toEqual({
+      iss: 'foo',
+      sub: 'bar',
+      aud: 'qux',
+      exp: 2000,
+      iat: 1000,
+    });
+  });
+
+  test('decoding invalid JWT string should throw Error', async () => {
+    expect(() => decodeIdToken('invalid-JWT')).toThrow('invalid token');
+  });
+
+  test('decoding valid JWT without iss should throw StructError', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const jwt = await new SignJWT({})
+      .setProtectedHeader({ alg: 'RS256' })
+      .setSubject('bar')
+      .setAudience('qux')
+      .setExpirationTime('2h')
+      .setIssuedAt()
+      .sign(privateKey);
+    expect(() => decodeIdToken(jwt)).toThrowError(StructError);
+  });
+});
--- a/packages/js/src/utils/id-token.ts
+++ b/packages/js/src/utils/id-token.ts
@@ -1,0 +1,53 @@
+import * as s from 'superstruct';
+
+const fullfillBase64 = (input: string) => {
+  if (input.length === 2) {
+    return `${input}==`;
+  }
+
+  if (input.length === 3) {
+    return `${input}=`;
+  }
+
+  return input;
+};
+
+/**
+ * @link [ID Token](https://openid.net/specs/openid-connect-core-1_0.html#IDToken)
+ */
+const IdTokenClaimsSchema = s.type({
+  iss: s.string(),
+  sub: s.string(),
+  aud: s.string(),
+  exp: s.number(),
+  iat: s.number(),
+  at_hash: s.optional(s.string()),
+});
+
+export type IdTokenClaims = s.Infer<typeof IdTokenClaimsSchema>;
+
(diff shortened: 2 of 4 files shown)


packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,68 @@
...
+  aud: s.string(),
+  exp: s.number(),
+  iat: s.number(),
+  auth_time: s.optional(s.number()),
+  nonce: s.optional(s.string()),
+  acr: s.optional(s.string()),
+  amr: s.optional(s.array(s.string())),
+  azp: s.optional(s.string()),
+  at_hash: s.optional(s.string()),
+  c_hash: s.optional(s.string()),
```
gao-sun, line 30: please follow our sdk convention to reflect the true state of Logto. those optional props can be added later

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,68 @@
...
+});
+
+export type IdTokenClaims = s.Infer<typeof IdTokenClaimsSchema>;
+
+/**
+ * Decode ID Token from JWT, without verifying.
+ *
+ * Verifying JWT requires fetching public key first,
+ * this can not be done in a sync function,
+ * in some cases, verifying is not necessary.
```
gao-sun, line 40: these comments mean nothing here

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,68 @@
...
+ * @param token JWT string
+ * @returns IDToken combined with JWT Claims.
```
gao-sun, line 42: no need for this comments until necessary

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,68 @@
...
+  }
+
+  const payloadString = payloadPart.replace(/-/g, '+').replace(/_/g, '/');
+  const json = decodeURIComponent(
+    escape(Buffer.from(fullfillBase64(payloadString), 'base64').toString())
+  );
+
+  try {
+    // Use Superstruct to validate the json type
+    const idToken = JSON.parse(json) as IdTokenClaims;
```
gao-sun, line 58: don't think we need `JSON.parse` when we already imported `superstruct` especially using `s.assert`

gao-sun: also, be careful with `JSON.parse` since it's extremely slow

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,68 @@
...
+
+  if (!payloadPart) {
+    throw new Error('invalid token');
+  }
+
+  const payloadString = payloadPart.replace(/-/g, '+').replace(/_/g, '/');
+  const json = decodeURIComponent(
+    escape(Buffer.from(fullfillBase64(payloadString), 'base64').toString())
+  );
+
```
gao-sun, line 55: why do we need decode JWT manually if we already imported `jose`? please see
https://github.com/panva/jose/blob/main/docs/functions/jwt_verify.jwtVerify.md#readme

packages/js/src/utils/id-token.test.ts
```diff
@@ -0,0 +1,36 @@
...
+    const jwt = await new SignJWT({})
+      .setProtectedHeader({ alg: 'RS256' })
+      .setIssuer('foo')
+      .setSubject('bar')
+      .setAudience('qux')
+      .setExpirationTime('2h')
+      .setIssuedAt()
+      .sign(privateKey);
+    const idTokenClaims = decodeIdToken(jwt);
+    expect(idTokenClaims.sub).toEqual('bar');
```
gao-sun, line 18: shall we just use equal function for object?

gao-sun (changes requested):

packages/js/src/utils/id-token.ts
IceHe: Removed.

packages/js/src/utils/id-token.ts
IceHe: Removed.

packages/js/src/utils/id-token.test.ts
IceHe: Updated, PTAL.

packages/js/src/utils/id-token.ts
IceHe: `jwtVerify` require key.
`decodeIdToken` have only one parameter `idToken` and it cannot get the key.

Reference: https://www.notion.so/silverhand/decodeIdToken-7b3b739c397a431882a2e41bac8782f1

packages/js/src/utils/id-token.ts
IceHe: `decodeIdToken` is used to decode an ID Token without verification.

Using SuperSturct.Struct to parse ID Token seems that it cannot do it without verification.

image

packages/js/src/utils/id-token.ts
IceHe: Should use `UnsecuredJWT`, which doesn't require key, instead.

Reference: https://github.com/panva/jose/blob/main/docs/classes/jwt_unsecured.UnsecuredJWT.md#readme

IceHe: `UnsecuredJWT` doesn't work here since 

> The UnsecuredJWT class is a utility for dealing with { "alg": "none" } Unsecured JWTs.

simeng-li: imageimage

IceHe:
cannot reopen after force pushing this branch.
please see #128.

PR #128, opened 2022-01-19
feat(js): decodeIdToken

IceHe:
## Summary

- Add utility function `decodeIdToken`
- Add basic type `IdTokenClaims`

Related #126 

## Linear Issue Reference

LOG-1355
LOG-1357

## Testing

- decode valid JWT → return claims
- decode invalid JWT → throw Error
- decode valid JWT without some claim (e.g. `iss`) → throw StructError

---
@logto-io/eng

### DIFF
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -19,7 +19,9 @@
   },
   "dependencies": {
     "@silverhand/essentials": "^1.1.3",
-    "js-base64": "^3.7.2"
+    "jose": "^4.1.4",
+    "js-base64": "^3.7.2",
+    "superstruct": "^0.15.3"
   },
   "devDependencies": {
     "@peculiar/webcrypto": "^1.1.7",
--- a/packages/js/src/utils/id-token.test.ts
+++ b/packages/js/src/utils/id-token.test.ts
@@ -1,0 +1,112 @@
+import { SignJWT, generateKeyPair } from 'jose';
+import { StructError } from 'superstruct';
+
+import { decodeIdToken } from './id-token';
+
+describe('decodeIdToken', () => {
+  test('decoding valid JWT should return claims', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const jwt = await new SignJWT({})
+      .setProtectedHeader({ alg: 'RS256' })
+      .setIssuer('foo')
+      .setSubject('bar')
+      .setAudience('qux')
+      .setExpirationTime(2000)
+      .setIssuedAt(1000)
+      .sign(privateKey);
+    const idTokenClaims = decodeIdToken(jwt);
+    expect(idTokenClaims).toEqual({
+      iss: 'foo',
+      sub: 'bar',
+      aud: 'qux',
+      exp: 2000,
+      iat: 1000,
+    });
+  });
+
+  test('decoding valid JWT with non-predefined claims should return all claims', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const payloadWithNonPredefinedClaims = { foo: 'bar' };
+    const jwt = await new SignJWT(payloadWithNonPredefinedClaims)
+      .setProtectedHeader({ alg: 'RS256' })
+      .setIssuer('foo')
+      .setSubject('bar')
+      .setAudience('qux')
+      .setExpirationTime(2000)
+      .setIssuedAt(1000)
+      .sign(privateKey);
+    const idTokenClaims = decodeIdToken(jwt);
+    expect(idTokenClaims).toEqual({
+      iss: 'foo',
+      sub: 'bar',
+      aud: 'qux',
+      exp: 2000,
+      iat: 1000,
+      foo: 'bar',
+    });
+  });
+
+  test('decoding invalid JWT string should throw Error', async () => {
+    expect(() => decodeIdToken('invalid-JWT')).toThrow('invalid token');
+  });
+
+  test('decoding valid JWT without issuer should throw StructError', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const jwt = await new SignJWT({})
+      .setProtectedHeader({ alg: 'RS256' })
+      .setSubject('bar')
+      .setAudience('qux')
+      .setExpirationTime('2h')
+      .setIssuedAt()
+      .sign(privateKey);
+    expect(() => decodeIdToken(jwt)).toThrowError(StructError);
+  });
+
+  test('decoding valid JWT without subject should throw StructError', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const jwt = await new SignJWT({})
+      .setProtectedHeader({ alg: 'RS256' })
+      .setIssuer('foo')
+      .setAudience('qux')
+      .setExpirationTime('2h')
+      .setIssuedAt()
+      .sign(privateKey);
+    expect(() => decodeIdToken(jwt)).toThrowError(StructError);
+  });
+
+  test('decoding valid JWT without audience should throw StructError', async () => {
+    const { privateKey } = await generateKeyPair('RS256');
+    const jwt = await new SignJWT({})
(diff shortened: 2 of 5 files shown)


packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,39 @@
...
+  aud: s.string(),
+  exp: s.number(),
+  iat: s.number(),
+  at_hash: s.optional(s.string()),
+});
+
+export type IdTokenClaims = s.Infer<typeof IdTokenClaimsSchema>;
+
+export const decodeIdToken = (token: string): IdTokenClaims => {
+  const { 1: encodedPayload } = token.split('.');
```
xiaoyijun, line 19: nit:
```js
const [ _, encodedPayload ] = token.split('.');
```

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,39 @@
...
+  sub: s.string(),
+  aud: s.string(),
+  exp: s.number(),
+  iat: s.number(),
+  at_hash: s.optional(s.string()),
+});
+
+export type IdTokenClaims = s.Infer<typeof IdTokenClaimsSchema>;
+
+export const decodeIdToken = (token: string): IdTokenClaims => {
```
xiaoyijun, line 18: Can we use the jose lib to decode ID token like @gao-sun mentioned in this [PR](https://github.com/logto-io/kotlin/pull/74#discussion_r787345337)?

gao-sun: @IceHe tested, unfortunately jose cannot do the job.

packages/js/src/utils/id-token.test.ts
```diff
@@ -0,0 +1,90 @@
...
+      .sign(privateKey);
+    const idTokenClaims = decodeIdToken(jwt);
+    expect(idTokenClaims).toEqual({
+      iss: 'foo',
+      sub: 'bar',
+      aud: 'qux',
+      exp: 2000,
+      iat: 1000,
+    });
+  });
```
gao-sun, line 25: as i mentioned we also need to test if the provider returns more claims than we defined

packages/js/src/utils/id-token.ts
IceHe: eslint doesn't recommend me to use code above.

image

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,39 @@
...
+    throw new Error('invalid token');
+  }
+
+  const json = UrlSafeBase64.decode(encodedPayload);
+
+  try {
+    // Use Superstruct to validate the json type
+    const idToken = JSON.parse(json) as IdTokenClaims;
+    s.assert(idToken, IdTokenClaimsSchema);
+    return idToken;
```
gao-sun, line 31: why do we still have `JSON.parse` here?

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,39 @@
...
+  } catch (error: unknown) {
+    if (error instanceof s.StructError) {
+      throw error;
+    }
+
+    throw new Error('invalid token: JSON parse failed');
+  }
```
gao-sun, line 38: no need for try - catch 😂

gao-sun (changes requested):

packages/js/src/utils/id-token.ts
IceHe: https://files.slack.com/files-pri/T0243NVUC9E-F02UZQYRQJV/image.png

IceHe: image

packages/js/src/utils/id-token.ts
IceHe: Removed, PTAL.

packages/js/src/utils/id-token.ts
IceHe: rewrote like this, PTAL:
```ts
  const json = UrlSafeBase64.decode(encodedPayload);
  return s.create(json, IdTokenClaimsSchema);
}
```

packages/js/src/utils/id-token.ts
```diff
@@ -0,0 +1,39 @@
...
+  at_hash: s.optional(s.string()),
+});
+
+export type IdTokenClaims = s.Infer<typeof IdTokenClaimsSchema>;
+
+export const decodeIdToken = (token: string): IdTokenClaims => {
+  const { 1: encodedPayload } = token.split('.');
+
+  if (!encodedPayload) {
+    throw new Error('invalid token');
```
simeng-li, line 22: Could refer to the Error definitions in Kotlin or Swift SDKs. Could define multiple type of errors for specific exceptions case. e.g. throw new JWTError('Invalid JWT format'). or throw new JWTError('Invalid token claims')

P.S Do we have a test case for this?

packages/js/src/utils/id-token.ts
simeng-li: I think the `json` value from `UrlSafeBase64.decode` is a string. Not sure we could bypass the JSON.parse.

packages/js/src/utils/id-token.ts
simeng-li: This is based on the previous comments. The try..catch might be valid if we have to keep JSON.parse and throw a controllable error type.

packages/js/src/utils/id-token.ts
IceHe: Still use `JSON.parse` according to JOSE implementation.

packages/js/src/utils/id-token.test.ts
IceHe: Added a new test case 'decoding valid JWT with non-predefined claims should return all claims', PTAL.

packages/js/src/utils/id-token.ts
IceHe: The variable named `json` can also present a string value containing json content. e.g. demo code from https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/JSON/parse

image

packages/js/src/utils/id-token.ts
IceHe: @wangsijie Any idea?

IceHe: Conclusion: No need for try-catch ( See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1642588712357800) )

packages/js/src/utils/id-token.ts
IceHe: Conclusion: Do not define custom Error for now, and do it when needed ( See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1642588826359200)

packages/js/package.json
```diff
@@ -19,7 +19,9 @@
   },
   "dependencies": {
     "@silverhand/essentials": "^1.1.2",
-    "js-base64": "^3.7.2"
+    "jose": "^4.1.4",
```
gao-sun, line 22: this should be put in `devDependencies` for now, or users will install `jose` while they don't need

packages/js/src/utils/id-token.ts
gao-sun: still need a custom error here. would you like to add a linear issue before merging?

gao-sun (approved):
overall lgtm, approve to unblock

IceHe:
> @gao-sun : still need a custom error here. would you like to add a linear issue before merging?

Added LOG-1355

packages/js/package.json
IceHe: My bad.

But it will be required by `verifyIdToken` in the next PR.
I prefer keep it for now.

packages/js/src/utils/id-token.ts
IceHe: Added LOG-1355

PR #131, opened 2022-01-21
chore(js): add LogtoErrorCode.idToken.invalidToken and jest-matcher-specific-error

IceHe:
## Summary

- Add `LogtoErrorCode.idToken.invalidToken`
    - Make decodeIdToken throw `LogtoErrorCode.idToken.invalidToken` when needed
- Simplify `LogtoErrorCode.idToken.verification.invalidIat` to `LogtoErrorCode.idToken.invalidIat`
- Add development dependency `jest-matcher-specific-error` for matching specific error

## Linear Issue Reference

LOG-1382
LOG-1355

## Testing

Pass all existing tests.

---
@logto-io/eng

PR #127, opened 2022-01-18
feat(js): verifyIdToken

IceHe:
## Summary

- Add `verifyIdToken`
    - Custom Error `LogtoError`

## Linear Issue Reference

LOG-1356

## Testing

- valid ID Token should not throw
- mismatched signature should throw
- mismatched issuer should throw
- mismatched audience should throw
- expired ID Token should throw
- issued at time, too far away from current time, should throw

---
@logto-io/eng

PR #133, opened 2022-01-21
refactor(js): assign LogtoError code to message

IceHe:
## Summary

refactor(js): assign LogtoError code to message too

---

Problem:

```ts
  test('callback uri, without code, should throw', () => {
    ……
    expect(() => ……).toThrowError(new LogtoError('callbackUri.missingCode'));
  });
```

is as same as

```ts
  test('callback uri, without code, should throw', () => {
    ……
    expect(() => ……).toThrowError();
  });
```

because

```ts
/**
 * If you want to test that a specific error is thrown inside a function.
 */
toThrowError(error?: string | Constructable | RegExp | Error): R;
```

`toThrowError` can only check the properties (e.g. `message`) defined in `Error`,
and cannot check the properties (e.g. `code`) defined in classes `extends` from `Error`.

My solution is to assign the property `LogtoError.code` to its parent class property `Error.message` 
so that `toThrowError` can assert the content of `LogtoError.code`.

## Linear Issue Reference

Related #127 #131 LOG-1355 LOG1356

## Testing

Pass all existing unit tests.

---
@logto-io/eng

### DIFF
--- a/packages/js/src/utils/errors.ts
+++ b/packages/js/src/utils/errors.ts
@@ -24,8 +24,8 @@ export type LogtoErrorCode = Normalize<typeof logtoErrorCodes>;
 export class LogtoError extends Error {
   code: LogtoErrorCode;
 
-  constructor(code: LogtoErrorCode, message?: string) {
-    super(message);
+  constructor(code: LogtoErrorCode) {
+    super(code);
     this.code = code;
   }
 }


gao-sun:
You can use `jest-matcher-specific-error` https://github.com/logto-io/logto/blob/master/packages/core/src/database/insert-into.test.ts

IceHe:
Follow the best practice.

PR #134, opened 2022-01-24
chore: add bootstrap script to post-install stage

demonzoo:
## Summary

add `lerna boostrap` script to post-install stage

## Testing

[x] At js monorepo root, execute `pnpm i`, the boostrap script is also executed at the end of the process

image


demonzoo:
Closing this PR as team said the bootstrap script should be triggered automatically, and thus not needed to put it in post-install.

PR #137, opened 2022-01-25
feat(js): requester

IceHe:
## Summary

- Add Requester for `fetch` Response
    - [`fetchOidcConfig`](https://www.notion.so/silverhand/fetchOidcConfig-887d329a10de43b5acc2f279c04dba4b) requires `Requester` to fetch data.
        - Reference: [packages/client/src/api/discover.ts](https://github.com/logto-io/js/blob/master/packages/client/src/api/discover.ts)
    - The developers should provide his/her `Requester` (`fetch` function) under Node.js environments according to the code implemented by @wangsijie before.
        - Reference: [packages/client/src/utils/requester.ts](https://github.com/logto-io/js/blob/master/packages/client/src/utils/requester.ts)

## Linear Issue Reference

LOG-766 → LOG-1421

## Testing

image
image

---
@logto-io/eng

### DIFF
--- a/packages/js/src/utils/errors.ts
+++ b/packages/js/src/utils/errors.ts
@@ -13,6 +13,10 @@ const logtoErrorCodes = Object.freeze({
     state_mismatched: 'State mismatched',
     missing_code: 'Missing code',
   },
+  requester: {
+    failed: 'Failed',
+    not_provide_fetch: 'Should provide a fetch function under Node.js',
+  },
 });
 
 export type LogtoErrorCode = NormalizeKeyPaths<typeof logtoErrorCodes>;
@@ -33,8 +37,8 @@ export class LogtoError extends Error {
   error?: string;
   errorDescription?: string;
 
-  constructor(code: LogtoErrorCode, error?: string, errorDescription?: string) {
-    super(getMessageByErrorCode(code));
+  constructor(code: LogtoErrorCode, error?: string, errorDescription?: string, message?: string) {
+    super(message ?? getMessageByErrorCode(code));
     this.code = code;
     this.error = error;
     this.errorDescription = errorDescription;
--- a/packages/js/src/utils/requester.node.test.ts
+++ b/packages/js/src/utils/requester.node.test.ts
@@ -1,0 +1,11 @@
+/**
+ * @jest-environment node
+ */
+import { LogtoError } from './errors';
+import { createRequester } from './requester';
+
+describe('createRequester', () => {
+  test('should throw when not providing fetch function under node.js environment', () => {
+    expect(() => createRequester()).toMatchError(new LogtoError('requester.not_provide_fetch'));
+  });
+});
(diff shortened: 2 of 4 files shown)


packages/js/src/utils/errors.ts
```diff
@@ -13,6 +13,10 @@ const logtoErrorCodes = Object.freeze({
     state_mismatched: 'State mismatched',
     missing_code: 'Missing code',
   },
+  requester: {
+    not_ok: 'Not ok',
+    not_provide_fetch: 'Should provide a fetch function under Node.js',
```
demonzoo, line 18: I don't quite get it, why do we have to emphasize "under Node.js"?
And let's put grammar aside, `A fetch function should be provided / is required` should be enough, IMO.

packages/js/src/utils/errors.ts
```diff
@@ -13,6 +13,10 @@ const logtoErrorCodes = Object.freeze({
     state_mismatched: 'State mismatched',
     missing_code: 'Missing code',
   },
+  requester: {
+    not_ok: 'Not ok',
```
demonzoo, line 17: not_ok => failed?

packages/js/src/utils/errors.ts
IceHe: Under browser environments, we use `window.fetch` to fetch response by default.
Under Node.js environments, the developers can use `node-fetch`, `got` or else.
Not provide default implementation under Node.js environments.
( follow the default implementation from @wangsijie )

Need further discussion.

packages/js/src/utils/errors.ts
IceHe: You mean
```
requester: {
    failed: 'Failed',
}
```
(I will fix it later)

demonzoo: Yeah, this looks better to me

packages/js/src/utils/errors.ts
demonzoo: OK, I think I see your points. We can discuss it offline tomorrow

packages/js/src/utils/errors.ts
IceHe: Updated, PTAL.

wangsijie (approved):
lgtm

packages/js/src/utils/errors.ts
demonzoo: Discussed, no problem for now. 
The current implementation requires the end user to specify a fetch function implementation in node.js environment. But for most of the cases, the environment would be the browser, so browser "fetch" function will be used by default.

packages/js/src/utils/errors.ts
```diff
@@ -33,8 +37,8 @@ export class LogtoError extends Error {
   error?: string;
   errorDescription?: string;
 
-  constructor(code: LogtoErrorCode, error?: string, errorDescription?: string) {
-    super(getMessageByErrorCode(code));
+  constructor(code: LogtoErrorCode, error?: string, errorDescription?: string, message?: string) {
```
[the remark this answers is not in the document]
IceHe: Resolve in another PR #139

PR #138, opened 2022-01-26
feat(js): fetchOidcConfig

IceHe:
## Summary

- Add core core function fetchOidcConfig
    - Add type OidcConfigResponse

## Linear Issue Reference

LOG-766

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/js/src/core/oidc-config.test.ts
+++ b/packages/js/src/core/oidc-config.test.ts
@@ -1,0 +1,26 @@
+import { fetchOidcConfig } from './oidc-config';
+
+describe('fetchOidcConfig', () => {
+  test('should return OidcConfigResponse with camel case keys', async () => {
+    const requester = jest.fn().mockResolvedValue({
+      authorization_endpoint: 'foo',
+      token_endpoint: 'foo',
+      userinfo_endpoint: 'foo',
+      end_session_endpoint: 'foo',
+      revocation_endpoint: 'foo',
+      jwks_uri: 'foo',
+      issuer: 'foo',
+    });
+    await expect(
+      fetchOidcConfig('https://example.com/oidc/.well-known/openid-configuration', requester)
+    ).resolves.toEqual({
+      authorizationEndpoint: 'foo',
+      tokenEndpoint: 'foo',
+      userinfoEndpoint: 'foo',
+      endSessionEndpoint: 'foo',
+      revocationEndpoint: 'foo',
+      jwksUri: 'foo',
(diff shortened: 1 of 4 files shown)


packages/js/src/core/oidc-config.ts
```diff
@@ -0,0 +1,16 @@
...
+  endSessionEndpoint: string;
+  revocationEndpoint: string;
+  jwksUri: string;
+  issuer: string;
+}
+
+export const fetchOidcConfig = async (
+  endpoint: string,
+  requester: Requester = createRequester()
+): Promise<OidcConfigResponse> => requester<OidcConfigResponse>(endpoint);
```
gao-sun, line 16: the return keys are snake_case

IceHe: Fixed, PTAL.

packages/js/src/core/oidc-config.test.ts
```diff
@@ -0,0 +1,26 @@
...
+      fetchOidcConfig('https://example.com/oidc/.well-known/openid-configuration', requester)
+    ).resolves.toEqual({
+      authorizationEndpoint: 'foo',
+      tokenEndpoint: 'foo',
+      userinfoEndpoint: 'foo',
+      endSessionEndpoint: 'foo',
+      revocationEndpoint: 'foo',
+      jwksUri: 'foo',
+      issuer: 'foo',
+    });
```
demonzoo, line 24: How about defining a mocked snake case key object, and use camelcase-keys to convert it to camel case key object? So that we can avoid creating these two objects twice.
```typescript
const  mockedOidcResponse = {
  authorization_endpoint: 'foo',
  token_endpoint: 'foo',
  userinfo_endpoint: 'foo',
  end_session_endpoint: 'foo',
  revocation_endpoint: 'foo',
  jwks_uri: 'foo',
  issuer: 'foo',
};

const requester = jest.fn().mockResolvedValue(mockedOidcResponse);

await expect(
  fetchOidcConfig('https://example.com/oidc/.well-known/openid-configuration', requester)
).resolves.toEqual(camelcaseKeys(mockedOidcResponse));
```

IceHe: If I use `camelcaseKeys` here too, both implementation and unit tests of `fetchOidcConfig` do the same thing (run same code).

And then we cannot reach the goal of testing the implementaion.

gao-sun (approved):

PR #141, opened 2022-01-26
feat(js): fetchTokenByRefreshToken

demonzoo:
## Summary

Implement js core function: fetchTokenByRefreshToken

## Linear Issue Reference

https://linear.app/silverhand/issue/LOG-771/fetch-tokens-by-refresh-token

## Testing

- [x] Passed all unit test cases

### DIFF
--- a/packages/js/src/consts/index.ts
+++ b/packages/js/src/consts/index.ts
@@ -1,0 +1,3 @@
+export const ContentType = {
+  formUrlEncoded: { 'Content-Type': 'application/x-www-form-urlencoded' },
+};
--- a/packages/js/src/core/fetch-token.test.ts
+++ b/packages/js/src/core/fetch-token.test.ts
@@ -1,0 +1,40 @@
+import { conditional, isNode } from '@silverhand/essentials';
+
+import { fetchTokenByRefreshToken, RefreshTokenTokenResponse } from './fetch-token';
+
+describe('fetch access token by providing valid refresh token', () => {
+  test('get token response', async () => {
+    const mockedOidcResponse = {
+      access_token: 'access_token',
+      refresh_token: 'refresh_token',
+      id_token: 'id_token',
+      scope: 'read register manage',
+      expires_in: 3600,
+    };
+
+    const expectedTokenResponse: RefreshTokenTokenResponse = {
+      accessToken: 'access_token',
+      refreshToken: 'refresh_token',
+      idToken: 'id_token',
+      scope: ['read', 'register', 'manage'],
+      expiresIn: 3600,
+    };
+
+    const fetchFunction = jest.fn().mockResolvedValue({
+      ok: true,
+      json: async () => mockedOidcResponse,
+    });
+
+    const tokenResponse = await fetchTokenByRefreshToken(
+      {
+        clientId: 'client_id',
+        tokenEndPoint: 'https://logto.dev/oidc/token',
+        refreshToken: 'refresh_token',
+        scope: ['read', 'register', 'manage'],
+      },
+      conditional(isNode() && fetchFunction)
+    );
+
+    expect(tokenResponse).toMatchObject(expectedTokenResponse);
+  });
+});
--- a/packages/js/src/core/fetch-token.ts
+++ b/packages/js/src/core/fetch-token.ts
@@ -1,0 +1,64 @@
+import { KeysToCamelCase } from '@silverhand/essentials';
+import { string, number, assert, type, optional, Infer } from 'superstruct';
+import { Except } from 'type-fest';
+
+import { ContentType } from '../consts';
+import { createRequester } from '../utils/requester';
+
+export interface FetchTokenByRefreshTokenParameters {
+  clientId: string;
+  tokenEndPoint: string;
+  refreshToken: string;
+  resource?: string;
+  scope?: string[];
+}
+
+const TokenResponseSchema = type({
+  access_token: string(),
+  refresh_token: string(),
+  id_token: string(),
+  scope: optional(string()),
+  expires_in: number(),
+});
+
+export type RefreshTokenTokenResponse = Except<
+  KeysToCamelCase<Infer<typeof TokenResponseSchema>>,
+  'scope'
+> & {
+  scope?: string[];
+};
+
+export const fetchTokenByRefreshToken = async (
+  { clientId, tokenEndPoint, refreshToken, resource, scope }: FetchTokenByRefreshTokenParameters,
+  fetchFunction?: typeof fetch
+): Promise<RefreshTokenTokenResponse> => {
+  const parameters = new URLSearchParams();
+  parameters.append('client_id', clientId);
+  parameters.append('refresh_token', refreshToken);
+
+  if (resource) {
+    parameters.append('resource', resource);
+  }
+
+  if (scope?.length) {
+    parameters.append('scope', scope.join(' '));
+  }
+
+  const requester = createRequester(fetchFunction);
+  const response = await requester(tokenEndPoint, {
+    method: 'POST',
+    headers: ContentType.formUrlEncoded,
+    body: parameters,
+  });
+
+  assert(response, TokenResponseSchema);
+  const { access_token, refresh_token, id_token, scope: response_scope, expires_in } = response;
+
+  return {
+    accessToken: access_token,
+    refreshToken: refresh_token,
+    idToken: id_token,
+    scope: response_scope?.split(' '),
+    expiresIn: expires_in,
+  };
+};
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -18,7 +18,7 @@
     "report": "WITH_REPORT=true npm run package"
   },
   "dependencies": {
-    "@silverhand/essentials": "^1.1.4",
+    "@silverhand/essentials": "^1.1.5",
     "jose": "^4.3.8",
     "js-base64": "^3.7.2",
     "lodash.get": "^4.4.2",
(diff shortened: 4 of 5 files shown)


packages/js/jest.config.js
```diff
@@ -3,6 +3,7 @@ module.exports = {
   coverageReporters: ['text-summary', 'lcov'],
   moduleFileExtensions: ['ts', 'js'],
   setupFilesAfterEnv: ['./jest.setup.js', 'jest-matcher-specific-error'],
+  testEnvironment: 'jsdom',
```
gao-sun, line 6: @IceHe shall we align the usage of setting environment? I saw file level declarations as well.

packages/js/src/core/fetch-token.ts
```diff
@@ -0,0 +1,75 @@
...
+  clientId,
+  tokenEndPoint,
+  refreshToken,
+  resource,
+  scope,
+}: FetchTokenByRefreshTokenParameters): Promise<RefreshTokenTokenResponse> => {
+  const parameters = new URLSearchParams();
+  parameters.append('client_id', clientId);
+  parameters.append('refresh_token', refreshToken);
+  if (resource) {
```
gao-sun, line 41: empty line before this line?

packages/js/src/core/fetch-token.ts
```diff
@@ -0,0 +1,75 @@
...
+    parameters.append('resource', resource);
+  }
+
+  if (scope?.length) {
+    parameters.append('scope', scope.join(' '));
+  }
+
+  const requester = createRequester();
```
gao-sun, line 49: Will it be better to use a shared requester instead of creating every time?

packages/js/src/core/fetch-token.ts
```diff
@@ -0,0 +1,75 @@
...
+    return {
+      accessToken: access_token,
+      refreshToken: refresh_token,
+      idToken: id_token,
+      scope: scope.split(' '),
+      expiresIn: expires_in,
+    };
+  } catch (error: unknown) {
+    if (error instanceof StructError) {
+      throw new LogtoError('struct_verification.response_struct_mismatched');
```
gao-sun, line 70: We can let superstruct throw the error directly instead of wrapping another layer to keep simplicity.

packages/playground/package.json
```diff
@@ -10,7 +10,7 @@
     "dev:razzle": "razzle start",
     "start": "concurrently -c \"blue,cyan\" \"pnpm:dev:tsc\" -k \"pnpm:dev:razzle\"",
     "start:prod": "NODE_ENV=production node build/server.js",
-    "build": "tsc -b && razzle build",
+    "build": "tsc -b && razzle build --noninteractive",
```
gao-sun, line 13: This should be submit in a separate PR, also not sure if we need this. I'm ok with both.

demonzoo: I think we do need this, as sometimes the CLI prompts for yes/no. And if we don't disable this, the CI will fail, which was exactly the case for me last night.

image

packages/js/src/core/fetch-token.ts
demonzoo: OK, fine for me. Will update shortly

packages/js/src/core/fetch-token.ts
demonzoo: OK, let me take a look at the linting rules

packages/js/jest.config.js
demonzoo: If I remember correctly, the file level declarations was just for the node environment. So declaring the default test environment as 'jsdom', and use file level declarations for node environment as an exception makes sense to me.

packages/js/src/core/fetch-token.ts
demonzoo: Like a singleton instance? If that's necessary, I think it would be better to be done in the `createRequester` itself

packages/js/jest.config.js
gao-sun: why do we need file level "node" declaration?

simeng-li: actually since js core SDK is not meant for dom only. we could leave it as default ?

packages/js/src/core/fetch-token.ts
gao-sun: Imaging we'd like to use some lib "foo" to send requests, then how do we achieve it with current design?

packages/playground/package.json
gao-sun: Interesting, why CI passed before?

packages/js/jest.config.js
demonzoo: Please take a look at [this discussion in PR#137](https://github.com/logto-io/js/pull/137/files#r791602491)
Looks like we have a node env specific error handling

packages/js/src/core/fetch-token.ts
```diff
@@ -0,0 +1,75 @@
...
+
+  if (scope?.length) {
+    parameters.append('scope', scope.join(' '));
+  }
+
+  const requester = createRequester();
+  const response = await requester(tokenEndPoint, {
+    method: 'POST',
+    headers: {
+      'Content-Type': 'application/x-www-form-urlencoded',
```
simeng-li, line 53: Actually I am thinking, as for the `content_type` header, should it be placed to some common place or encapsulated within the requester itself by default?  So no need for us to provide it manually in every api.

packages/js/src/core/fetch-token.ts
```diff
@@ -0,0 +1,75 @@
...
+    },
+    body: parameters,
+  });
+
+  try {
+    assert(response, TokenResponseSchema);
+    const { access_token, refresh_token, id_token, scope, expires_in } = response;
```
simeng-li, line 60: nit: `{ access_token: accessToken, refresh_token: refreshToken ... } = response; `

packages/js/src/core/fetch-token.ts
```diff
@@ -0,0 +1,75 @@
...
+  refreshToken: string;
+  resource?: string;
+  scope?: string[];
+}
+
+const TokenResponseSchema = type({
+  access_token: string(),
+  refresh_token: string(),
+  id_token: string(),
+  scope: string(),
```
simeng-li, line 20: I am not quite sure whether the scope return value is required or not. If no scope if provided, I believe it could be undefined. Let's go with the optional for now. I'll update the doc.

packages/playground/package.json
demonzoo: No idea. I would also like to know why. 😂
Anyways, I will try to remove this change from this PR, and open up a new one.

packages/js/src/core/fetch-token.ts
demonzoo: ok, makes sense to me

packages/js/src/core/fetch-token.ts
demonzoo: I agree. Should I put it in a `js/consts/index.ts` file?

packages/js/jest.config.js
wangsijie: js core SDK can be run in node environment.

packages/js/src/core/fetch-token.ts
demonzoo: Done.

packages/js/src/core/fetch-token.ts
demonzoo: I tried, but personally don't quite like the outcome. For me it does NOT seem to increase readability.

image

packages/js/src/core/fetch-token.ts
demonzoo: Uh-huh, I see your points. Code updated to allow user to pass in a fetchFunction. Please let me know if this solves the problem. Thx.

packages/playground/package.json
demonzoo: Interesting enough, I reverted the changes made to disable the interactive build, but this time the CI is working...
Not sure what went wrong earlier, but now it seems the changes will no longer be needed, at least for now

demonzoo: @gao-sun you can resolve this comment now. The code change has been reverted.

simeng-li (approved):
LGTM

packages/js/src/consts/index.ts
```diff
@@ -0,0 +1,3 @@
+export const CONTENT_TYPE = {
+  FORM_URL_ENCODED: { 'Content-Type': 'application/x-www-form-urlencoded' },
```
gao-sun, line 2: we don't have a convention of using upper snake case for constants. Just use camelCase for the key and PascalCase for the object name is fine.

packages/js/src/core/fetch-token.test.ts
```diff
@@ -0,0 +1,36 @@
...
+  test('get token response', async () => {
+    const mockedOidcResponse = {
+      access_token: 'access_token',
+      refresh_token: 'refresh_token',
+      id_token: 'id_token',
+      scope: 'read register manage',
+      expires_in: 3600,
+    };
+
+    const expectedTokenResponse: RefreshTokenTokenResponse = {
```
gao-sun, line 15: use camelcaseKeys？

gao-sun (approved):
overall looks good, please resolve conversation before merging

packages/js/jest.config.js
demonzoo: After offline team discussion, we decided to run tests for both jsdom and node envs, by executing the `run test` command twice in parallel in CI. Thus remove this env setting from jest.config.js

packages/js/src/core/fetch-token.test.ts
demonzoo: `camelcaseKeys` or `camelCaseKeys`?

packages/js/src/consts/index.ts
demonzoo: done

PR #146, opened 2022-01-28
test(js): add test script for jsdom

gao-sun:
## Summary

add jsdom test script to `test:coverage` in `js`. not using GitHub Actions matrix since it's for one package only.

## Testing

UT

## Reviewers
@logto-io/eng

### DIFF
--- a/packages/js/src/core/fetch-token.test.ts
+++ b/packages/js/src/core/fetch-token.test.ts
@@ -1,5 +1,3 @@
-import { conditional, isNode } from '@silverhand/essentials';
-
 import { fetchTokenByRefreshToken, RefreshTokenTokenResponse } from './fetch-token';
 
 describe('fetch access token by providing valid refresh token', () => {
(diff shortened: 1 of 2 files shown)


packages/js/src/core/fetch-token.test.ts
```diff
@@ -32,7 +30,7 @@ describe('fetch access token by providing valid refresh token', () => {
         refreshToken: 'refresh_token',
         scope: ['read', 'register', 'manage'],
       },
-      conditional(isNode() && fetchFunction)
+      fetchFunction // Always passing `fetchFunction` since Jest has no `fetch()`
```
demonzoo, line 33: But if we run it in jsdom env, there is a fetch implementation by browser, isn't it?

gao-sun: no, jest is running in node so there's no `fetch` until we use a headless browser environment.

demonzoo: If you specify --env=jsdom, then there IS a fetch implementation, right?

demonzoo: OK, just googled it and I'm definetely wrong. Sorry about that 😂

demonzoo (approved):

PR #147, opened 2022-01-28
feat(js): refactor fetchTokenByRefreshToken

IceHe:
## Summary

- simplify fetchTokenByRefreshToken with camelcaseKeys
- remove useless superstruct.assert in fetchTokenByRefreshToken

## Linear Issue Reference

LOG-782

## Testing

Pass all existing tests.

---
@logto-io/eng

### DIFF
--- a/packages/js/src/core/fetch-token.ts
+++ b/packages/js/src/core/fetch-token.ts
@@ -1,6 +1,5 @@
 import { KeysToCamelCase } from '@silverhand/essentials';
 import camelcaseKeys from 'camelcase-keys';
-import { string, number, assert, type, optional, Infer } from 'superstruct';
 import { Except } from 'type-fest';
 
 import { ContentType } from '../consts';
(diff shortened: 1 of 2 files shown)


packages/js/src/core/fetch-token.ts
```diff
@@ -45,20 +42,13 @@ export const fetchTokenByRefreshToken = async (
...
   const requester = createRequester(fetchFunction);
-  const response = await requester(tokenEndPoint, {
+  const tokenSnakeCaseResponse = await requester<TokenSnakeCaseResponse>(tokenEndPoint, {
     method: 'POST',
     headers: ContentType.formUrlEncoded,
     body: parameters,
   });
 
-  assert(response, TokenResponseSchema);
-  const { access_token, refresh_token, id_token, scope: response_scope, expires_in } = response;
```
demonzoo, line 55: I think we do have to keep the assert? Otherwise if the actual response fails to meet our typescript types in runtime, we don't have a guard to prevent it from causing further issues

demonzoo: OK, based on the latest team discussion, we can assume the server response is always in correct type, as it is totally controllable. Thus the assert is not necessary any more.

IceHe: Unnecessary. Logto Core server is under our control, and we assume its response will have  the correct structure or the code should fail directly.

demonzoo (approved):

packages/js/src/core/fetch-token.ts
```diff
@@ -13,20 +13,17 @@ export interface FetchTokenByRefreshTokenParameters {
   scope?: string[];
```
gao-sun, line 13: this should be non-nullable

IceHe: according to SDK Convention, this input parameter should be optional

wangsijie (approved):
LGTM

PR #149, opened 2022-01-29
feat(js): generateSignInUri and scope

IceHe:
## Summary

- feat(js): scope
- feat(js): generateSignInUri

## Linear Issue Reference

LOG-768

## Testing

image
image

---
@logto-io/eng

### DIFF
--- a/packages/js/src/core/sign-in.ts
+++ b/packages/js/src/core/sign-in.ts
@@ -1,0 +1,57 @@
+import qs from 'query-string';
+
+import { withDefaultScopeValues } from '../utils/scope';
+
+const prompt = 'consent';
+const codeChallengeMethod = 'S256';
+
+export type SignInUriParameters = {
+  authorizationEndpoint: string;
+  clientId: string;
+  redirectUri: string;
+  codeChallenge: string;
+  state: string;
+  scope?: string | string[];
+  resource?: string | string[];
+};
+
+enum ResponseType {
+  code = 'code',
+}
+
+const toArray = (stringOrStringArray?: string | string[]): string[] =>
+  stringOrStringArray === undefined
+    ? []
+    : Array.isArray(stringOrStringArray)
+    ? stringOrStringArray
+    : [stringOrStringArray];
+
+export const generateSignInUri = ({
+  authorizationEndpoint,
+  clientId,
+  redirectUri,
+  codeChallenge,
+  state,
+  scope,
+  resource,
+}: SignInUriParameters) => {
+  const queryStringWithoutResource = qs.stringify({
+    client_id: clientId,
+    redirect_uri: [redirectUri],
+    code_challenge: codeChallenge,
+    code_challenge_method: codeChallengeMethod,
+    state,
+    response_type: ResponseType.code,
+    prompt,
+    scope: withDefaultScopeValues(scope),
+  });
+
+  const queryStringWithOnlyResource = toArray(resource).reduce(
+    (accumulator, item) => `${accumulator}&resource=${item}`,
+    ''
+  );
+
+  const queryString = `${queryStringWithoutResource}${queryStringWithOnlyResource}`;
+
+  return `${authorizationEndpoint}?${queryString}`;
+};
--- a/packages/js/src/utils/scope.test.ts
+++ b/packages/js/src/utils/scope.test.ts
@@ -1,0 +1,82 @@
+import { defaultScopeString, ScopeValue, withDefaultScopeValues } from './scope';
+
+describe('withDefaultScopeValues', () => {
+  describe('should append default scope values to the head of scope string', () => {
+    describe('undefined param', () => {
+      test('should return DEFAULT_SCOPE_STRING if calling without parameters', () => {
+        expect(withDefaultScopeValues()).toEqual(defaultScopeString);
+      });
+    });
+
+    describe('string param', () => {
+      test('empty', () => {
+        expect(withDefaultScopeValues('')).toEqual(defaultScopeString);
+      });
+
+      test('with blank characters', () => {
+        expect(withDefaultScopeValues(' \t')).toEqual(defaultScopeString);
+      });
+
+      test('with openid offline_access', () => {
+        expect(withDefaultScopeValues(defaultScopeString)).toEqual(defaultScopeString);
+      });
+
+      test('with openid (without default scope value offline_access)', () => {
+        expect(withDefaultScopeValues(ScopeValue.openid)).toEqual(defaultScopeString);
+      });
+
+      test('with openid name (without default scope value offline_access)', () => {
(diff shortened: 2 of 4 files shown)


packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,59 @@
...
+    redirect_uri: [redirectUri],
+    code_challenge: codeChallenge,
+    code_challenge_method: 'S256',
+    state,
+    response_type: ResponseType.code,
+    prompt: 'consent',
+    scope: joinStrings(scope),
+  });
+
+  const queryStringWithOnlyResource = wrapInStringArray(resource).reduce(
```
xiaoyijun, line 51: Maybe we can limit the `scope` type as `string[]`.
[Slack Discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1643439717664529)

packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,59 @@
...
+    : stringOrStringArray;
+
+const wrapInStringArray = (stringOrStringArray?: string | string[]): string[] =>
+  stringOrStringArray === undefined
+    ? []
+    : Array.isArray(stringOrStringArray)
+    ? stringOrStringArray
+    : [stringOrStringArray];
+
+export const generateSignInUri = ({
```
xiaoyijun, line 31: per [discussion oflline](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1642498392221400) and convention we should ensure `openid` & `offline_access` is contained by scope.

xiaoyijun (changes requested):

packages/js/src/core/sign-in.ts
IceHe: Fixed, PTAL.

packages/js/src/core/sign-in.ts
IceHe: Simplified, PTAL

packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,47 @@
...
+    redirect_uri: [redirectUri],
+    code_challenge: codeChallenge,
+    code_challenge_method: 'S256',
+    state,
+    response_type: ResponseType.code,
+    prompt: 'consent',
+    scope: withDefaultScopes(scope),
+  });
+
+  const queryStringWithOnlyResource = (resource ?? []).reduce(
```
demonzoo, line 39: `resource?.reduce`?

packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,47 @@
...
+  redirectUri: string;
+  codeChallenge: string;
+  state: string;
+  scope?: string | string[];
+  resource?: string[];
+};
+
+enum ResponseType {
+  code = 'code',
+}
```
demonzoo, line 17: Do we have other response type options in the future?

packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,47 @@
...
+  resource,
+}: SignInUriParameters) => {
+  const queryStringWithoutResource = qs.stringify({
+    client_id: clientId,
+    redirect_uri: [redirectUri],
+    code_challenge: codeChallenge,
+    code_challenge_method: 'S256',
+    state,
+    response_type: ResponseType.code,
+    prompt: 'consent',
```
demonzoo, line 35: Do we need to have enums for `code_challenge_method` and `prompt`?

packages/js/src/core/sign-in.ts
IceHe: Yes, add when needed. prefer not to pre-define.

packages/js/src/utils/scope.test.ts
```diff
@@ -0,0 +1,82 @@
+import { DEFAULT_SCOPE_STRING, ScopeValue, withDefaultScopes } from './scope';
+
+describe('withDefaultScopes', () => {
```
demonzoo, line 3: Should be better if we use natural words instead of camel case. E.g. `Generate sign-in URI with default scopes`

packages/js/src/utils/scope.test.ts
```diff
@@ -0,0 +1,82 @@
...
+  describe('Expect withDefaultScopes method to append default scope values to the head of scope string', () => {
+    describe('undefined param', () => {
```
demonzoo, line 5: Why do we have nested `describe` blocks?

packages/js/src/utils/scope.test.ts
IceHe: i'm used to naming the top level description with the name of function i test.
i'm ok with both.

packages/js/src/utils/scope.test.ts
IceHe: group the similar test cases

packages/js/src/utils/scope.ts
```diff
@@ -0,0 +1,26 @@
+export enum ScopeValue {
+  name = 'name',
+  email = 'email',
+  offline_access = 'offline_access',
+  openid = 'openid',
+}
+
+export const DEFAULT_SCOPE_VALUES = [ScopeValue.openid, ScopeValue.offline_access];
+export const DEFAULT_SCOPE_STRING = DEFAULT_SCOPE_VALUES.join(' ');
```
demonzoo, line 9: I think @gao-sun once said we don't have convension to declare upper snake case variables, would be just fine to use camel case here

packages/js/src/core/sign-in.ts
IceHe: can not
image

xiaoyijun (dismissed):

IceHe:
The branch name is wrong.
I will close it and create new PR after fixing all comments.

packages/js/src/core/sign-in.ts
demonzoo: OK, I missed that one. So you will have to either fallback the array (`resource ?? []`) or the result string (`?? ''`). Then I'm OK with this.

demonzoo:
> The branch name is wrong (related to wrong linear issue) . I will close it and create new PR after fixing all comments.

Alright..

packages/js/src/core/sign-in.ts
IceHe: added local constants for now and will extract them as enum when needed.

packages/js/src/utils/scope.ts
IceHe: - upper snake case for global constants
- camel case for local constants

IceHe: OK, based on the latest team discussion, we should use camel case name for all constants for js sdk

IceHe:
New PR #150 with branch icehe-log-768

PR #150, opened 2022-01-29
feat(js): scopes and generateSignInUri

IceHe:
## Summary

- feat(js): scopes
- feat(js): generateSignInUri

---
Related PR: #149 

## Linear Issue Reference

LOG-768

## Testing

image
image

---
@logto-io/eng

### DIFF
--- a/packages/js/src/core/sign-in.ts
+++ b/packages/js/src/core/sign-in.ts
@@ -1,0 +1,38 @@
+import qs from 'query-string';
+
+import { withReservedScopes } from '../utils/scopes';
+
+const codeChallengeMethod = 'S256';
+const prompt = 'consent';
+const responseType = 'authorization_code';
+
+export type SignInUriParameters = {
+  authorizationEndpoint: string;
+  clientId: string;
+  redirectUri: string;
+  codeChallenge: string;
+  state: string;
+  scopes?: string[];
+  resources?: string[];
+};
+
+export const generateSignInUri = ({
+  authorizationEndpoint,
+  clientId,
+  redirectUri,
+  codeChallenge,
+  state,
+  scopes,
+  resources,
+}: SignInUriParameters) =>
+  `${authorizationEndpoint}?${qs.stringify({
+    client_id: clientId,
+    redirect_uri: redirectUri,
+    code_challenge: codeChallenge,
+    code_challenge_method: codeChallengeMethod,
+    state,
+    response_type: responseType,
+    prompt,
+    scope: withReservedScopes(scopes),
+    resource: resources ?? [],
+  })}`;
--- a/packages/js/src/utils/scopes.test.ts
+++ b/packages/js/src/utils/scopes.test.ts
@@ -1,0 +1,38 @@
+import { withReservedScopes } from './scopes';
+
+const email = 'email';
+const name = 'name';
+const offlineAccess = 'offline_access';
+const openid = 'openid';
+const reservedScopes = [openid, offlineAccess];
+const reservedScopesString = reservedScopes.join(' ');
+
+describe('withReservedScopes', () => {
+  test('with undefined param', () => {
+    expect(withReservedScopes()).toEqual(reservedScopesString);
+  });
+
+  test('with nothing', () => {
+    expect(withReservedScopes([])).toEqual(reservedScopesString);
+  });
+
+  test('with all reserved scopes, openid and offline_access', () => {
+    expect(withReservedScopes([openid, offlineAccess])).toEqual(reservedScopesString);
+  });
+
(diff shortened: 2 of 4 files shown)


packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,50 @@
...
+  clientId,
+  redirectUri,
+  codeChallenge,
+  state,
+  scopes,
+  resources,
+}: SignInUriParameters) => {
+  const queryStringWithoutResources = qs.stringify({
+    client_id: clientId,
+    redirect_uri: [redirectUri],
```
demonzoo, line 33: Why is redirect_uri an array?

IceHe: Fixed, PTAL

packages/js/src/utils/scopes.ts
```diff
@@ -0,0 +1,9 @@
+/**
+ * @param originalScopes
+ * @return scopes should contain all reserved scopes ( Logto requires `openid` and `offline_access` )
+ */
+export const withReservedScopes = (originalScopes?: string[]): string => {
+  const uniqueScopes = new Set(['openid', 'offline_access', ...(originalScopes ?? [])]);
+
+  return Array.from(uniqueScopes).slice().sort().join(' ');
```
demonzoo, line 8: What does the `slice()` do? Shallow copy the array?
I think `Array.from()` already does the shallow copy, so it's not necessary to `slice()` again

packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,50 @@
...
+    prompt,
+    scope: withReservedScopes(scopes),
+  });
+
+  const queryStringWithOnlyResources = (resources ?? []).reduce(
+    (accumulator, item) => `${accumulator}&resource=${item}`,
+    ''
+  );
+
+  const queryString = `${queryStringWithoutResources}${queryStringWithOnlyResources}`;
```
simeng-li, line 47: why not just use qs.stringfy a list value?

packages/js/src/core/sign-in.ts
```diff
@@ -0,0 +1,50 @@
+import qs from 'query-string';
+
+import { withReservedScopes } from '../utils/scopes';
+
+const prompt = 'consent';
+const codeChallengeMethod = 'S256';
+
+enum ResponseType {
+  code = 'code',
+}
```
demonzoo, line 10: So why is response type defined as an enum, while the other two above are constants? Use constants for all?

packages/js/src/utils/scopes.test.ts
```diff
@@ -0,0 +1,38 @@
+import { withReservedScopes } from './scopes';
+
+const email = 'email';
+const name = 'name';
+const offlineAccess = 'offline_access';
+const openid = 'openid';
```
demonzoo, line 6: nit: `const openId = 'open_id'`?

gao-sun: @demonzoo `openid` is reserved scope in oidc, please check rfc

demonzoo: ok, my bad

packages/js/src/utils/scopes.ts
gao-sun: it's lint requirement to prevent unexpected inline array sorting

demonzoo: I see. It is to avoid direct mutation

demonzoo:
Hey, the test cases are failed. Please take a look

packages/js/src/utils/scopes.ts
IceHe: `sort` is unnecessary. will remove `sort`

packages/js/src/core/sign-in.ts
IceHe: My bad

packages/js/src/core/sign-in.ts
IceHe: - `prompt` is not suitable to define as enum for now
- `codeChallengeMethod` is fixed (defined by RFC)
- since more response types will be used in the near future, i prefer to predefine `ResponseType` (but define the only used one for now) (will be exported and moved to suitable place)

demonzoo: How about adding it when we really need it? Just took a look at swift sdk and found Gao was using string variable.

IceHe: Updated, PTAL.

demonzoo (approved):
LGTM

wangsijie (approved):
LGTM

PR #152, opened 2022-01-30
feat(js): fetchUserInfo

IceHe:
## Summary

feat(js): fetchUserInfo

## Linear Issue Reference

LOG-1124

## Testing

image

---
@eng

### DIFF
--- a/packages/js/src/core/user-info.test.ts
+++ b/packages/js/src/core/user-info.test.ts
@@ -1,0 +1,14 @@
+import { fetchUserInfo, UserInfoResponse } from './user-info';
+
+describe('fetchUserInfo', () => {
+  test('should return UserInfoResponse', async () => {
+    const userInfoResponse: UserInfoResponse = { sub: 'foo' };
+    const fetchFunction = jest.fn().mockResolvedValue({
+      ok: true,
+      json: async () => userInfoResponse,
+    });
+    await expect(
+      fetchUserInfo('https://example.com/oidc/me', 'access_token_value', fetchFunction)
+    ).resolves.toMatchObject(userInfoResponse);
+  });
+});
--- a/packages/js/src/core/user-info.ts
+++ b/packages/js/src/core/user-info.ts
@@ -1,0 +1,17 @@
+import { createRequester } from '../utils/requester';
+
+export type UserInfoResponse = {
+  sub: string;
+};
+
+export const fetchUserInfo = async (
+  userInfoEndpoint: string,
+  accessToken: string,
+  fetchFunction?: typeof fetch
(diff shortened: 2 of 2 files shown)


packages/js/src/core/user-info.ts
```diff
@@ -0,0 +1,14 @@
+import { createRequester, Requester } from '../utils/requester';
+
+export type UserInfoResponse = {
+  sub: string;
```
simeng-li, line 4: only sub being returned?

simeng-li (changes requested):

packages/js/src/core/user-info.test.ts
```diff
@@ -0,0 +1,11 @@
+import { fetchUserInfo, UserInfoResponse } from './user-info';
+
+describe('fetchUserInfo', () => {
+  test('should return UserInfoResponse', async () => {
+    const userInfoResponse: UserInfoResponse = { sub: 'foo' };
+    const requester = jest.fn().mockResolvedValue(userInfoResponse);
+    await expect(
+      fetchUserInfo('https://example.com/oidc/me', 'access_token_value', requester)
+    ).resolves.toEqual(userInfoResponse);
```
wangsijie, line 9: `toMatchObject`?

wangsijie (approved):
overall LGTM

packages/js/src/core/user-info.ts
IceHe: yes, only sub for now
https://www.notion.so/silverhand/oidc-me-3a281ed217894558be0651e983f9ac2c

packages/js/src/core/user-info.test.ts
IceHe: I'm ok with both. 
Name of `toMatchObject` seems to be more precise.
Updated, PTAL.

packages/js/src/core/user-info.ts
```diff
@@ -0,0 +1,17 @@
...
+};
+
+export const fetchUserInfo = async (
+  userInfoEndpoint: string,
+  accessToken: string,
+  fetchFunction?: typeof fetch
+): Promise<UserInfoResponse> => {
```
demonzoo, line 11: I'm wondering if we should define all core function params as object pattern, instead of directly pass in the params, since we already have that pattern in other core functions (e.g. sign-in, sign-out).

IceHe: Prefer to not define type `XxxParameters` until needed

IceHe: Based on the discussion on Slack, just use simple parameters.

simeng-li (approved):

PR #160, opened 2022-02-07
feat(browser): add LogtoClient constructor

IceHe:
## Summary

- feat(browser): add LogtoClient constructor
- test(browser): construct LogtoClient

## Linear Issue Reference

LOG-786

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/browser/jest.config.js
+++ b/packages/browser/jest.config.js
@@ -2,6 +2,8 @@ module.exports = {
   coveragePathIgnorePatterns: ['/node_modules/', '/lib/'],
   coverageReporters: ['text-summary', 'lcov'],
   moduleFileExtensions: ['ts', 'js'],
+  setupFilesAfterEnv: ['./jest.setup.js'],
+  testEnvironment: 'jsdom',
   testPathIgnorePatterns: ['lib'],
   testRegex: '(/__tests__/.*|\\.(test|spec))\\.(ts|js)$',
   transform: {
--- a/packages/browser/jest.setup.js
+++ b/packages/browser/jest.setup.js
@@ -1,0 +1,8 @@
+// Need to disable following rules to mock text-decode/text-encoder and crypto for jsdom
+// https://github.com/jsdom/jsdom/issues/1612
+/* eslint-disable @silverhand/fp/no-mutation */
+/* eslint-disable unicorn/prefer-module */
+const { TextDecoder, TextEncoder } = require('text-encoder');
+
+global.TextDecoder = TextDecoder;
+global.TextEncoder = TextEncoder;
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -28,6 +28,7 @@
     "jest": "^27.0.6",
     "lint-staged": "^12.3.3",
     "prettier": "^2.3.2",
+    "text-encoder": "^0.0.4",
     "ts-jest": "^27.0.4",
     "ts-loader": "^9.2.6",
     "typescript": "^4.3.5",
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -1,8 +1,17 @@
-import { TODO } from './index';
+import { createRequester } from '@logto/js';
 
-describe('TODO', () => {
-  test('TODO', () => {
-    const todo: TODO = 'TODO';
-    expect(todo).toEqual('TODO');
+import LogtoClient from './index';
+
+describe('LogtoClient', () => {
+  test('constructor', () => {
+    expect(
+      () =>
+        new LogtoClient({
+          endpoint: 'https://logto.dev',
+          clientId: 'client_id_value',
+          usingPersistStorage: false,
+          requester: createRequester(),
+        })
+    ).not.toThrow();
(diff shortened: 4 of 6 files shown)


packages/browser/src/modules/logto-client.test.ts
```diff
@@ -0,0 +1,19 @@
...
+import LogtoClient from './logto-client';
+
+describe('LogtoClient', () => {
+  test('constructor', () => {
+    expect(
+      () =>
+        new LogtoClient({
+          endpoint: 'https://logto.dev/oidc/.well-known/openid-configuration',
+          clientId: 'client_id_value',
+          scopes: ['openid', 'offline_access'],
```
gao-sun, line 12: `openid` and `offline_access` should be reserved scopes thus no need to pass when create a client. actually we can hide the param for now

packages/browser/src/modules/logto-client.ts
```diff
@@ -0,0 +1,28 @@
+import { OidcConfigResponse, Requester } from '@logto/js';
+
+export type LogtoConfig = {
+  endpoint: string;
+  clientId: string;
+  scopes?: string[];
+  resources?: string[];
+  usingPersistStorage: false;
```
gao-sun, line 8: should be boolean?

packages/browser/src/modules/logto-client.ts
```diff
@@ -0,0 +1,28 @@
...
+  requester: Requester;
+};
+
+export type AccessToken = {
+  token: string;
+  scope: string;
+  expiresAt: number; // In seconds
```
gao-sun, line 15: not in seconds? this should be unix timestamp

packages/browser/src/modules/logto-client.ts
IceHe: I just follow the SDK convention. Maybe need discussion?
image

packages/browser/src/modules/logto-client.ts
IceHe: I just follow SDK Convention. 

Status quo:

- Kotlin SDK uses seconds.

image

- Swift SDK uses milliseconds.

image


Need to align.

image

packages/browser/src/modules/logto-client.test.ts
```diff
@@ -0,0 +1,19 @@
+import { createRequester } from '@logto/js';
+
+import LogtoClient from './logto-client';
+
+describe('LogtoClient', () => {
+  test('constructor', () => {
+    expect(
+      () =>
+        new LogtoClient({
+          endpoint: 'https://logto.dev/oidc/.well-known/openid-configuration',
```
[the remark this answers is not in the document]
IceHe: Fixed. PTAL @gao-sun

packages/browser/src/modules/logto-client.test.ts
IceHe: Updated. PTAL @gao-sun

packages/browser/src/modules/logto-client.ts
simeng-li: @xiaoyijun 错了吧，应该是timestamp

simeng-li: ？

IceHe: Fixed. PTAL @gao-sun

packages/browser/src/modules/logto-client.ts
demonzoo: But the type should be boolean, with `false` being the default value of that boolean property.
So I think it should be `optional` as well, as only optional property can have default value. 

Thus:
```
usingPersistStorage?: boolean;
```

demonzoo (approved):

packages/browser/src/modules/logto-client.ts
IceHe: Updated. PTAL @gao-sun @demonzoo

packages/browser/src/modules/logto-client.ts
```diff
@@ -0,0 +1,28 @@
...
+  protected accessTokenMap = new Map<string, AccessToken>();
+  protected refreshToken?: string;
+  protected idToken?: string;
+  protected logtoConfig: LogtoConfig;
+  protected oidcConfig?: OidcConfigResponse;
+
+  constructor(logtoConfig: LogtoConfig) {
+    this.logtoConfig = logtoConfig;
+  }
+}
```
demonzoo, line 28: Wait, I thought this would be the `index.ts` of the browser package, but now I realized it's not. Why wouldn't you put this LogtoClient class in the index?

IceHe: u'r right. better with `index.ts` right now.

xiaoyijun (approved):
Overall LGTM

gao-sun (approved):

packages/browser/src/modules/logto-client.ts
IceHe: Finally // Unix TimeStamp in seconds

PR #155, opened 2022-01-30
chore(js): override eslint config to allow 6 params at max

demonzoo:
## Summary

Override eslint config in project `logto-io/js` to allow 6 function parameters at maximum

## Testing

- [x] Eslint will prompt a warning message when declaring over 6 parameters in any function
image

PR #161, opened 2022-02-07
chore(js): remove eslint override of max-params

demonzoo:
## Summary

Revert the changes made in #155 as we decided to use the object pattern in the end.
Thus, remove the override of eslint rule of `max-params` in js project

## Testing

No need


IceHe (approved):
LGTM

PR #162, opened 2022-02-07
refactor(js): shorten imports and replace interface with type

IceHe:
## Summary

- chore(js): shorten imports
- refactor(js): replace interface with type
    - Reference: https://www.typescriptlang.org/docs/handbook/2/everyday-types.html#differences-between-type-aliases-and-interfaces
    - Discussion: https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644231317277599

## Linear Issue Reference

None.

## Testing

Pass all existing tests

---
@logto-io/eng

### DIFF
--- a/packages/js/src/core/fetch-token.ts
+++ b/packages/js/src/core/fetch-token.ts
@@ -3,7 +3,7 @@ import camelcaseKeys from 'camelcase-keys';
 import { Except } from 'type-fest';
 
 import { ContentType } from '../consts';
-import { Requester } from '../utils/requester';
+import { Requester } from '../utils';
 
 export interface FetchTokenByRefreshTokenParameters {
   clientId: string;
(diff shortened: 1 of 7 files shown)


demonzoo:
This is more like a `refactor` to me, not `chore`

packages/js/src/core/fetch-token.ts
```diff
@@ -3,15 +3,15 @@ import camelcaseKeys from 'camelcase-keys';
 import { Except } from 'type-fest';
 
 import { ContentType } from '../consts';
-import { Requester } from '../utils/requester';
+import { Requester } from '../utils';
 
-export interface FetchTokenByRefreshTokenParameters {
```
demonzoo, line 8: Have we decided to use `type` instead of `interface` in all cases?

IceHe:
> This is more like a refactor to me, not chore

Updated, PTAL @demonzoo

demonzoo (approved):

gao-sun (approved):

PR #166, opened 2022-02-08
refactor(js): replace query-string with URLSearchParams and refactor callback-uri.ts

IceHe:
## Summary

- [refactor(js): replace query-string with URLSearchParams](https://github.com/logto-io/js/commit/e09214f8c62814cdb437b4c358b79af2fb183ddc)
- [refactor(js): callback-uri.ts](https://github.com/logto-io/js/commit/0084479657746635b3b91addeb761a3df3378b37)

## Linear Issue Reference

LOG-1460

## Testing

Pass all existing tests.

---
@logto-io/eng


packages/js/src/core/sign-in.ts
```diff
@@ -34,5 +32,11 @@ export const generateSignInUri = ({
...
     scope: withReservedScopes(scopes),
-    resource: resources ?? [],
-  })}`;
+  });
+
+  for (const resource of resources ?? []) {
+    urlSearchParameters.append('resource', resource);
+  }
+
+  return `${authorizationEndpoint}?${urlSearchParameters.toString()}`;
```
demonzoo, line 41: What if the resources is undefined? The current code logic will always return the endpoint with the `?` appended in the end, regardless of the resources existence.

packages/js/src/utils/callback-uri.ts
```diff
@@ -23,19 +17,20 @@ export const verifyAndParseCodeFromCallbackUri = (
...
 
-  const authenticationResult = parseCallbackUri(callbackUri);
-  const {
-    code,
-    state: stateFromCallbackUri,
-    error,
-    error_description: errorDescription,
-  } = authenticationResult;
+  const error = conditional(uriParameters.get(QueryKey.Error));
+  const errorDescription = conditional(uriParameters.get(QueryKey.ErrorDescription));
```
gao-sun, line 23: why do we need `conditional` here?

IceHe: My bad. There is no need to do it now.

(Did it for some nit reason that is invalid now.)

IceHe: Fixed. PTAL @gao-sun .

gao-sun (approved):

PR #169, opened 2022-02-10
feat(browser): isAuthenticated getter method

demonzoo:
## Summary

Implemented `isAuthenticated` getter method in LogtoClient.
We consider the user is authenticated when we have gained/loaded the ID token successfully.


gao-sun (changes requested):
the head branch is wrong

demonzoo:
> the head branch is wrong

It was on purpose, as I implemented the code based on that. Will rebase once #170 is merged.

wangsijie (approved):

PR #170, opened 2022-02-10
feat(browser): signIn

IceHe:
## Summary

Main:

- feat(browser): signIn
- test(browser): signIn

---

Chores:

- chore(js,browser): add exports in package.json
    - Fix webpack failure: [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644473115488029) on Slack
- chore(js,browser): replace npm with pnpm in package.json
- feat(js): export consts
- feat(js): export parseUriParameters

---  

Next up:

- `handleCallback`

## Linear Issue Reference

LOG-790

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/browser/jest.setup.js
+++ b/packages/browser/jest.setup.js
@@ -2,7 +2,15 @@
 // https://github.com/jsdom/jsdom/issues/1612
 /* eslint-disable @silverhand/fp/no-mutation */
 /* eslint-disable unicorn/prefer-module */
+const crypto = require('crypto');
+
+const { location } = require('jest-location-mock');
 const { TextDecoder, TextEncoder } = require('text-encoder');
 
+global.crypto = {
+  getRandomValues: (buffer) => crypto.randomFillSync(buffer),
+  subtle: crypto.webcrypto.subtle,
+};
+global.location = location;
 global.TextDecoder = TextDecoder;
 global.TextEncoder = TextEncoder;
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -28,6 +28,7 @@
     "codecov": "^3.8.3",
     "eslint": "^8.8.0",
     "jest": "^27.0.6",
+    "jest-location-mock": "^1.0.9",
     "lint-staged": "^12.3.3",
     "prettier": "^2.3.2",
     "text-encoder": "^0.0.4",
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -19,7 +19,8 @@
     "report": "WITH_REPORT=true pnpm package"
   },
   "dependencies": {
-    "@logto/js": "^0.0.1"
+    "@logto/js": "^0.0.1",
+    "@silverhand/essentials": "^1.1.5"
   },
   "devDependencies": {
     "@silverhand/eslint-config": "^0.7.0",
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -15,8 +15,8 @@
     "test": "jest",
     "test:coverage": "jest --silent --coverage",
     "test:report": "codecov -F browser",
-    "prepublish": "npm run build && npm run package",
-    "report": "WITH_REPORT=true npm run package"
+    "prepublish": "pnpm build && pnpm package",
+    "report": "WITH_REPORT=true pnpm package"
   },
   "dependencies": {
     "@logto/js": "^0.0.1"
--- a/packages/browser/package.json
+++ b/packages/browser/package.json
@@ -2,6 +2,7 @@
   "name": "@logto/browser",
   "version": "0.0.1",
   "main": "lib/index.js",
+  "exports": "./lib/index.js",
   "typings": "lib/index.d.ts",
   "license": "MPL-2.0",
   "scripts": {
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -1,17 +1,30 @@
-import { createRequester } from '@logto/js';
-
 import LogtoClient from './index';
+
+const clientId = 'client_id_value';
+const endpoint = 'https://logto.dev';
+const authorizationEndpoint = `${endpoint}/oidc/auth`;
+const redirectUri = 'http://localhost:3000/callback';
+const requester = jest.fn();
+const signInUri = `${authorizationEndpoint}?foo=bar`;
+
+jest.mock('@logto/js', () => {
+  return {
+    ...jest.requireActual('@logto/js'),
+    fetchOidcConfig: async () => ({ authorizationEndpoint }),
+    generateSignInUri: () => signInUri,
+  };
+});
 
 describe('LogtoClient', () => {
   test('constructor', () => {
-    expect(
-      () =>
-        new LogtoClient({
-          endpoint: 'https://logto.dev',
-          clientId: 'client_id_value',
-          usingPersistStorage: false,
-          requester: createRequester(),
-        })
-    ).not.toThrow();
+    expect(() => new LogtoClient({ endpoint, clientId, requester })).not.toThrow();
+  });
+
+  describe('signIn', () => {
+    test('window.location should be correct signInUri', async () => {
+      const logtoClient = new LogtoClient({ endpoint, clientId, requester });
+      await logtoClient.signIn(redirectUri);
+      expect(window.location.toString()).toEqual(signInUri);
+    });
   });
 });
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -1,4 +1,18 @@
-import { OidcConfigResponse, Requester } from '@logto/js';
+import {
+  fetchOidcConfig,
+  generateCodeChallenge,
+  generateCodeVerifier,
+  generateSignInUri,
+  generateState,
+  OidcConfigResponse,
+  Requester,
+  withReservedScopes,
+} from '@logto/js';
+
+const discoveryPath = '/oidc/.well-known/openid-configuration';
+const logtoStorageItemKeyPrefix = `logto`;
+
+export const getLogtoKey = (key: string) => `${logtoStorageItemKeyPrefix}:${key}`;
 
 export type LogtoConfig = {
   endpoint: string;
@@ -24,5 +38,39 @@ export default class LogtoClient {
 
   constructor(logtoConfig: LogtoConfig) {
     this.logtoConfig = logtoConfig;
+  }
+
+  public async signIn(redirectUri: string) {
+    const { clientId, resources, scopes: customScopes } = this.logtoConfig;
+    const oidcConfig = await this.getOidcConfig();
+
+    const { authorizationEndpoint } = oidcConfig;
+    const codeVerifier = generateCodeVerifier();
+    const codeChallenge = await generateCodeChallenge(codeVerifier);
+    const state = generateState();
+    const scopes = withReservedScopes(customScopes).split(' ');
+
+    const signInUri = generateSignInUri({
+      authorizationEndpoint,
+      clientId,
+      redirectUri,
+      codeChallenge,
+      state,
+      scopes,
+      resources,
+    });
+
+    // TODO: save redirectUri, codeVerifier and state
+    window.location.assign(signInUri);
+  }
+
+  protected async getOidcConfig(): Promise<OidcConfigResponse> {
+    if (!this.oidcConfig) {
+      const { endpoint, requester } = this.logtoConfig;
+      const discoveryEndpoint = new URL(discoveryPath, endpoint);
+      this.oidcConfig = await fetchOidcConfig(discoveryEndpoint.toString(), requester);
+    }
+
+    return this.oidcConfig;
   }
 }
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -15,8 +15,8 @@
     "test": "jest",
     "test:coverage": "jest --silent --env=jsdom && jest --silent --coverage",
     "test:report": "codecov -F js",
-    "prepublish": "npm run build && npm run package",
-    "report": "WITH_REPORT=true npm run package"
+    "prepublish": "pnpm build && pnpm package",
+    "report": "WITH_REPORT=true pnpm package"
   },
   "dependencies": {
     "@silverhand/essentials": "^1.1.5",
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -2,6 +2,7 @@
   "name": "@logto/js",
   "version": "0.0.1",
   "main": "lib/index.js",
+  "exports": "./lib/index.js",
   "typings": "lib/index.d.ts",
   "license": "MPL-2.0",
   "scripts": {
--- a/packages/js/src/utils/callback-uri.ts
+++ b/packages/js/src/utils/callback-uri.ts
@@ -1,7 +1,7 @@
 import { QueryKey } from '../consts';
 import { LogtoError } from './errors';
 
-const parseUriParameters = (uri: string) => {
+export const parseUriParameters = (uri: string) => {
   const [, queryString = ''] = uri.split('?');
 
   return new URLSearchParams(queryString);
--- a/packages/js/src/index.ts
+++ b/packages/js/src/index.ts
@@ -1,2 +1,3 @@
 export * from './core';
 export * from './utils';
+export * from './consts';
--- a/pnpm-lock.yaml
+++ b/pnpm-lock.yaml
@@ -22,6 +22,7 @@ importers:
     specifiers:
       '@logto/js': ^0.0.1
       '@silverhand/eslint-config': ^0.7.0
+      '@silverhand/essentials': ^1.1.5
       '@silverhand/ts-config': ^0.7.0
       codecov: ^3.8.3
       eslint: ^8.8.0
@@ -37,6 +38,7 @@ importers:
       webpack-cli: ^4.9.1
     dependencies:
       '@logto/js': link:../js
+      '@silverhand/essentials': 1.1.5
     devDependencies:
       '@silverhand/eslint-config': 0.7.0_93c6ce1132bfea333665fcd5223669e0
       '@silverhand/ts-config': 0.7.0_typescript@4.4.4
@@ -4422,7 +4424,7 @@ packages:
       webpack-cli: 4.x.x
     dependencies:
       envinfo: 7.8.1
-      webpack-cli: 4.9.1_51b97b2ad784f46f7873debea239ad34
+      webpack-cli: 4.9.1_ffb4ca0d42777c3f95985c39e374aaf9
     dev: true
 
   /@webpack-cli/serve/1.6.0_webpack-cli@4.9.1:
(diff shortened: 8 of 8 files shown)


packages/browser/src/index.test.ts
```diff
@@ -1,17 +1,50 @@
...
+  });
+
+  describe('signInWithBrowser', () => {
+    test('sessionStorage and signInUri should contain correct properties', async () => {
+      const logtoClient = new LogtoClient({ endpoint, clientId, requester });
+      const originalRedirectUri = 'http://localhost:3000/callback';
+      await logtoClient.signInWithBrowser(originalRedirectUri);
+
+      const item = sessionStorage.getItem(getLogtoKey(clientId)) ?? '{}';
+      const logtoSignInSessionItem = JSON.parse(item) as LogtoSignInSessionItem;
```
demonzoo, line 29: This is wrong, as the item could be `{}`, so the parsed object `logtoSignInSessionItem` is likely to be an empty object `{}`, which doesn't match the type `LogtoSignInSessionItem`

packages/browser/src/index.test.ts
```diff
@@ -1,17 +1,50 @@
...
+
+      const state = logtoSignInSessionItem?.state;
+      const codeVerifier = logtoSignInSessionItem?.codeVerifier;
+      const codeChallenge = await generateCodeChallenge(codeVerifier);
+      const redirectUri = logtoSignInSessionItem?.redirectUri;
```
demonzoo, line 34: And based on your current code, the `logtoSignInSessionItem` is forced to be an instance of `LogtoSignInSessionItem`, and it will never be optional. Thus the `?` makes no sense in these sentences.
In fact, we could improve our linter to auto-fix these problems

packages/browser/src/index.test.ts
IceHe: I use `?? '{}'` and `as LogtoSignInSessionItem` on purpose.

If the item is null, `?? '{}'` will make logtoSignInSessionItem an empty object and  won't fail immediately.
And then I use the following `expect` statements (which can give a better message for the failed test case) to check whether sessionStorage contains the expected item with the expected properties.

`as LogtoSignInSessionItem` makes (e.g.) `logtoSignInSessionItem?.state` access legally (passing eslint).

Need some suggestions to help achieve these purposes for better UTs. @demonzoo

packages/js/package.json
```diff
@@ -14,8 +15,8 @@
     "test": "jest",
     "test:coverage": "jest --silent --env=jsdom && jest --silent --coverage",
     "test:report": "codecov -F js",
-    "prepublish": "npm run build && npm run package",
-    "report": "WITH_REPORT=true npm run package"
+    "prepublish": "pnpm run build && pnpm run package",
```
wangsijie, line 18: seems original `npm` works fine?

IceHe: yes. Update it due to [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644474687361579?thread_ts=1644473115.488029&cid=C02A8G4HVAM) on Slack.
I'm good with both (npm and pnpm) here.

Any better idea?

demonzoo: I'll go with `pnpm` if you ask me.  ;-)

packages/browser/src/index.test.ts
IceHe: Improved, @demonzoo PTAL

packages/js/package.json
IceHe: Finally `"prepublish": "pnpm package"`.
See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644549967093349) on Slack

packages/browser/src/index.test.ts
IceHe: Improved. @demonzoo  PTAL.

packages/browser/jest.setup.js
```diff
@@ -2,7 +2,13 @@
 // https://github.com/jsdom/jsdom/issues/1612
 /* eslint-disable @silverhand/fp/no-mutation */
 /* eslint-disable unicorn/prefer-module */
+const { Crypto } = require('@peculiar/webcrypto');
```
gao-sun, line 5: Will one of the following work:
1. Set jest env browser
2. See attached

image

packages/browser/jest.setup.js
```diff
@@ -2,7 +2,13 @@
...
+const { location } = require('jest-location-mock');
+const fetch = require('node-fetch');
 const { TextDecoder, TextEncoder } = require('text-encoder');
 
+global.crypto = new Crypto();
+global.location = location;
+global.fetch = fetch;
```
gao-sun, line 12: We do we need fetch in unit test?

gao-sun (changes requested):

packages/browser/jest.setup.js
IceHe: createRequester/Requester uses fetch.

packages/browser/jest.setup.js
IceHe: @gao-sun

1. jest-environment `jsdom` does not work.
   See https://github.com/jsdom/jsdom/issues/1612 jsdom open issue: window.crypto is undefined
2. It's one of the existing solution. I'm good with both.

image

cc: @wangsijie

packages/browser/jest.setup.js
IceHe: Removed node-fecth and improved UT. @gao-sun PTAL.

IceHe: Improved. @gao-sun PTAL.
```ts
const crypto = require('crypto');

……

global.crypto = {
  getRandomValues: (buffer) => crypto.randomFillSync(buffer),
  subtle: crypto.webcrypto.subtle,
};
```

packages/browser/jest.setup.js
IceHe: Improved.
```ts
const crypto = require('crypto');

……

global.crypto = {
  getRandomValues: (buffer) => crypto.randomFillSync(buffer),
  subtle: crypto.webcrypto.subtle,
};
```

demonzoo (approved):
LGTM

packages/browser/src/index.ts
```diff
@@ -25,4 +44,36 @@ export default class LogtoClient {
   constructor(logtoConfig: LogtoConfig) {
     this.logtoConfig = logtoConfig;
   }
+
+  public async signInWithBrowser(redirectUri: string) {
```
demonzoo, line 48: Do we need to specificly call this signIn`WithBrowser`?

IceHe: Follow SDK Convention https://www.notion.so/silverhand/signIn-bb0b307176f644be8f3e763189f00159

image

packages/browser/src/index.test.ts
```diff
@@ -1,17 +1,51 @@
...
+      expect(item).not.toBeNull();
+
+      if (item) {
+        const logtoSignInSessionItem = JSON.parse(item) as LogtoSignInSessionItem;
+        const { state, codeVerifier, redirectUri } = logtoSignInSessionItem;
+        const codeChallenge = await generateCodeChallenge(codeVerifier);
+
+        expect(redirectUri).toEqual(originalRedirectUri);
+        expect(state.length).toBeGreaterThan(1);
+        expect(codeVerifier.length).toBeGreaterThan(1);
```
gao-sun, line 38: nit: something like `.toBeGreaterThan(1).toLessOrEqualTo(128)`?

packages/browser/src/index.ts
gao-sun: we are already in a browser, so no need to append `WithBrowser`?

packages/browser/src/index.ts
```diff
@@ -25,4 +44,36 @@ export default class LogtoClient {
...
+      clientId,
+      redirectUri,
+      codeChallenge,
+      state,
+      scopes,
+      resources,
+    });
+
+    const logtoKey = getLogtoKey(clientId);
+    const logtoSignInSessionItem: LogtoSignInSessionItem = { redirectUri, codeVerifier, state };
```
gao-sun, line 73: we can also cache oidc config in session storage (let's add an issue and do it later)

packages/js/src/utils/callback-uri.ts
```diff
@@ -1,7 +1,7 @@
 import { QueryKey } from '../consts';
 import { LogtoError } from './errors';
 
-const parseUriParameters = (uri: string) => {
+export const parseUriParameters = (uri: string) => {
   const [, queryString = ''] = uri.split('?');
 
   return new URLSearchParams(queryString);
```
gao-sun, line 7: per [MDN](https://developer.mozilla.org/en-US/docs/Web/API/URLSearchParams), this class is web-worker only. let's add an issue and double check it later

packages/browser/src/index.ts
```diff
@@ -25,4 +44,36 @@ export default class LogtoClient {
   constructor(logtoConfig: LogtoConfig) {
     this.logtoConfig = logtoConfig;
   }
+
+  public async signInWithBrowser(redirectUri: string) {
+    const { clientId, endpoint, requester, resources, scopes: originalScopes } = this.logtoConfig;
+
+    this.oidcConfig = await fetchOidcConfig(
+      `${endpoint}/oidc/.well-known/openid-configuration`,
```
gao-sun, line 52: if we already used `URLSearchParams` then we can use [URL](https://developer.mozilla.org/en-US/docs/Web/API/URL/URL) for safe join too

gao-sun (approved):

packages/js/src/utils/callback-uri.ts
demonzoo: I'm afraid not, just checked Node.js documentation, and it's been there since v7.5
https://nodejs.org/api/url.html#class-urlsearchparams

packages/browser/src/index.test.ts
IceHe: seems to be better test these cases (length<=128) for `generateState` and `generateCodeVerifier` in `js/js`?
(I just intend to test they are not blank here.)

packages/browser/src/index.ts
IceHe: Simplified.

packages/browser/src/index.ts
IceHe: Improved. PTAL @gao-sun 

```ts

const discoveryPath = '/oidc/.well-known/openid-configuration';

……

  public async signIn(redirectUri: string) {
    ……
    const discoveryEndpoint = new URL(discoveryPath, endpoint);
    this.oidcConfig = await fetchOidcConfig(discoveryEndpoint.toString(), requester);
    ……

```

packages/browser/src/index.ts
IceHe: Added the linear issue LOG-1519

packages/browser/src/index.test.ts
gao-sun: then we should not assert anything here? since they all should be covered in js unit tests

packages/browser/src/index.ts
```diff
@@ -1,4 +1,24 @@
...
+  generateState,
+  OidcConfigResponse,
+  Requester,
+  withReservedScopes,
+} from '@logto/js';
+
+const discoveryPath = '/oidc/.well-known/openid-configuration';
+const logtoStorageItemKeyPrefix = `logto:`;
+
+export const getLogtoKey = (key: string) => `${logtoStorageItemKeyPrefix}${key}`;
```
simeng-li, line 15: would ${prefix}:${key} is cleaner?  more informable that colon is used to slipt the key and prefix?

packages/browser/src/index.ts
```diff
@@ -25,4 +45,33 @@ export default class LogtoClient {
...
+      clientId,
+      redirectUri,
+      codeChallenge,
+      state,
+      scopes,
+      resources,
+    });
+
+    const logtoKey = getLogtoKey(clientId);
+    const logtoSignInSessionItem: LogtoSignInSessionItem = { redirectUri, codeVerifier, state };
```
simeng-li, line 71: nit: just make it inline? 

```js
const logtoValue = JSON.stringify({ redirectUri, codeVerifier, state });
```

packages/browser/src/index.ts
```diff
@@ -25,4 +45,33 @@ export default class LogtoClient {
...
+    const logtoValue = JSON.stringify(logtoSignInSessionItem);
+    sessionStorage.setItem(logtoKey, logtoValue);
```
simeng-li, line 73: prefer to have a sessionStorage utiles to handle the set and get logic.  Wrap up all those, keyGenerate, JSONStringfy, JSON parse logic within the util method. simply call `logtoSesstion.set(),` `logtoSessiont.get()`

packages/js/src/utils/callback-uri.ts
simeng-li: https://caniuse.com/?search=URLSearchParams

packages/browser/src/index.test.ts
```diff
@@ -1,17 +1,51 @@
...
+
+        const signInUri = window.location.toString();
+        expect(signInUri.startsWith(authorizationEndpoint));
+
+        const urlSearchParameters = parseUriParameters(signInUri);
+        expect(urlSearchParameters.get(QueryKey.ClientId)).toEqual(clientId);
+        expect(urlSearchParameters.get(QueryKey.RedirectUri)).toEqual(originalRedirectUri);
+        expect(urlSearchParameters.get(QueryKey.CodeChallenge)).toEqual(codeChallenge);
+        expect(urlSearchParameters.get(QueryKey.State)).toEqual(state);
```
simeng-li, line 47: better split into two different test cases. 
1. generate signInUrl
2. session storage

packages/browser/src/index.ts
IceHe: intend to use type LogtoSignInSessionItem for better access when signIn and handleCallback.
better keep it？@simeng-li

packages/browser/src/index.ts
```diff
@@ -25,4 +45,33 @@ export default class LogtoClient {
   constructor(logtoConfig: LogtoConfig) {
     this.logtoConfig = logtoConfig;
   }
+
+  public async signIn(redirectUri: string) {
+    const { clientId, endpoint, requester, resources, scopes: originalScopes } = this.logtoConfig;
+    const discoveryEndpoint = new URL(discoveryPath, endpoint);
+    this.oidcConfig = await fetchOidcConfig(discoveryEndpoint.toString(), requester);
```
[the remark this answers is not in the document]
IceHe: I prefer to fetch OIDC config at first.
And I'm also good with fetch OIDC config when needed.
Finally i decided to conform to the majority,
since Swift & Kotlin SDK fetch OIDC config when needed (`oidcConfig` property is optional in LogtoClient) for now.

@logto-io/eng any idea?

packages/browser/src/index.test.ts
IceHe: Finally

```ts
        expect(codeVerifier.length).toBeGreaterThan(1);
        expect(codeVerifier.length).toBeLessThanOrEqual(128);
```

since `expect(…).toBeXXX(…)` calling cannot be chained.

packages/js/src/utils/callback-uri.ts
IceHe: Conclusion: It's safe to use URLSearchParams.
It is not Web Workers only API.

See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644809199979179) on Slack.

packages/browser/src/index.ts
IceHe: Canceled issue since it's reasonable to just cache oidc config only in object property.

See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644809320215889) on Slack.

packages/browser/src/index.ts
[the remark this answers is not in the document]
IceHe: Updated.

```ts
  public async signIn(redirectUri: string) {
    const { clientId, resources, scopes: customScopes } = this.logtoConfig;
    const oidcConfig = await this.getOidcConfig();

    const { authorizationEndpoint } = oidcConfig;
    ……
  }

  private async getOidcConfig() {
    if (!this.oidcConfig) {
      const { endpoint, requester } = this.logtoConfig;
      const discoveryEndpoint = new URL(discoveryPath, endpoint);
      this.oidcConfig = await fetchOidcConfig(discoveryEndpoint.toString(), requester);
    }

    return this.oidcConfig;
  }
```

See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644809320215889) .

packages/browser/jest.setup.js
```diff
@@ -2,7 +2,15 @@
 // https://github.com/jsdom/jsdom/issues/1612
 /* eslint-disable @silverhand/fp/no-mutation */
 /* eslint-disable unicorn/prefer-module */
```
demonzoo, line 4: Should provide ending block to form a disable-enable-pair

IceHe: Will fix in the next PR.

packages/browser/src/index.ts
IceHe: Do it in another way. PTAL.

packages/browser/src/index.ts
IceHe: extract two class methods to get and set the sign-in session item. PTAL.

See [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1644814026202089) on Slack.

packages/browser/package.json
```diff
@@ -14,18 +15,21 @@
...
-    "prepublish": "npm run build && npm run package",
-    "report": "WITH_REPORT=true npm run package"
+    "prepublish": "pnpm build && pnpm package",
+    "report": "WITH_REPORT=true pnpm package"
   },
   "dependencies": {
-    "@logto/js": "^0.0.1"
+    "@silverhand/essentials": "^1.1.5",
+    "@logto/js": "^0.0.1",
+    "superstruct": "^0.15.3"
```
gao-sun, line 24: no need for `superstruct` here?

gao-sun (approved):

packages/browser/src/index.test.ts
IceHe: The original code no longer exists.
Resolved this comment.

packages/js/src/core/oidc-config.ts
```diff
@@ -3,7 +3,7 @@ import camelcaseKeys from 'camelcase-keys';
 
 import { Requester } from '../utils';
 
-type OidcConfigSnakeCaseResponse = {
+export type OidcConfigSnakeCaseResponse = {
```
demonzoo, line 6: Is this necessary?

simeng-li (approved):
LGTM

packages/browser/src/index.test.ts
IceHe: The source code no longer exists and I resolve this commement.

packages/browser/src/index.ts
IceHe: Do it in another PR next up.

demonzoo (approved):
LGTM

packages/browser/src/index.ts
IceHe: The source code no longer exists and I resolve this comment.

PR #171, opened 2022-02-11
chore: bump configs to v0.7.0

charIeszhao:
## Summary

Bump configs to v0.7.0


xiaoyijun (approved):
LGTM

packages/react/src/reducer.ts
```diff
@@ -1,58 +1,49 @@
...
-  }
+type Action = {
+  type:
+    | 'INITIALIZE'
+    | 'LOGIN_WITH_REDIRECT'
+    | 'HANDLE_CALLBACK_REQUEST'
+    | 'HANDLE_CALLBACK_SUCCESS'
+    | 'LOGOUT'
+    | 'ERROR';
+  payload?: Partial<AuthState>;
```
wangsijie, line 11: The original type anotation is more specific?

wangsijie (changes requested):

packages/react/src/reducer.ts
demonzoo: Yes, it's more specific, but using the `no-unnecessary-condition` lint rule, the linter complained about the last `if condition` being `unnecessary`, thus I refactored the code to use switch-case instead.
But later the linter complained about lexical declarations in switch case (enabled by `no-case-declarations` rule), so I have to bring those declarations to the very top of the function to make the linter happy.

So I have to use a more generic type definition to make the variable declarations work at top level.

If we prefer using more specific type definitions for each `case`, we'll probably have to define several new `payload types` for each case, which I think is a bit off the track and unnecessary. So finally I came up with the idea of using a more generic type definition for `payload`.

wangsijie (approved):
Approve to unblock merging. Feel free to resolve the conversation after offline discussion (with a conclusion).

packages/react/src/reducer.ts
demonzoo: Well, I decided to revert the change but instead add a ignore for that last if-condition to avoid unnecessary headache

PR #174, opened 2022-02-14
feat(browser): sign out

demonzoo:
## Summary

Provide sign out method in browser sdk

## Linear Issue Reference

LOG-793

## Testing

Test cases will be provided in another PR that links to LOG-1245

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -3,9 +3,11 @@ import {
   generateCodeChallenge,
   generateCodeVerifier,
   generateSignInUri,
+  generateSignOutUri,
   generateState,
   OidcConfigResponse,
   Requester,
+  revoke,
   withReservedScopes,
 } from '@logto/js';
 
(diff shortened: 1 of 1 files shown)


packages/browser/src/index.ts
```diff
@@ -68,6 +70,34 @@ export default class LogtoClient {
     window.location.assign(signInUri);
   }
 
+  public async signOut(postLogoutRedirectUri?: string) {
+    if (!this.idToken || !this.oidcConfig) {
+      return;
```
wangsijie, line 75: maybe throw an error?

demonzoo: browser package doesn't have error handling utilities at this point, will throw error once #175 is merged.

wangsijie (approved):

packages/browser/src/index.ts
```diff
@@ -68,6 +70,34 @@ export default class LogtoClient {
...
+      void revoke(revocationEndpoint, clientId, this.refreshToken, requester);
+    }
+
+    const url = generateSignOutUri({
+      endSessionEndpoint,
+      postLogoutRedirectUri,
+      idToken: this.idToken,
+    });
+
+    window.location.assign(url);
```
simeng-li, line 92: nit: `window.location.assign(url);` could leave it as the last statement.

simeng-li (approved):
LGTM

packages/browser/src/index.ts
demonzoo: Code updated.

packages/browser/src/index.ts
gao-sun: should to await oidc config instead of directly returning.

packages/browser/src/index.ts
```diff
@@ -68,6 +70,35 @@ export default class LogtoClient {
...
+    if (!this.idToken || !this.oidcConfig) {
+      return;
+    }
+
+    const { clientId, requester } = this.logtoConfig;
+    const { endSessionEndpoint, revocationEndpoint } = this.oidcConfig;
+    const logtoKey = getLogtoKey(clientId);
+
+    if (this.refreshToken) {
+      void revoke(revocationEndpoint, clientId, this.refreshToken, requester);
```
gao-sun, line 83: this is not good since the request will be interrupted when `window.location` changed

PR #175, opened 2022-02-14
feat(browser): sign-in session storage

IceHe:
## Summary

feat(browser): sign-in session storage

Related #170 

## Linear Issue Reference

LOG-1544
LOG-790
LOG-882

## Testing

image

---
@logto-io/eng

### DIFF
--- a/packages/browser/src/errors.ts
+++ b/packages/browser/src/errors.ts
@@ -1,0 +1,33 @@
+import { NormalizeKeyPaths } from '@silverhand/essentials';
+import get from 'lodash.get';
+
+const logtoClientErrorCodes = Object.freeze({
+  sign_in_session: {
+    invalid: 'Invalid sign-in session.',
+  },
+});
+
+export type LogtoClientErrorCode = NormalizeKeyPaths<typeof logtoClientErrorCodes>;
+
+const getMessageByErrorCode = (errorCode: LogtoClientErrorCode): string => {
+  // TODO: linear issue LOG-1419
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const message = get(logtoClientErrorCodes, errorCode);
+
+  if (typeof message === 'string') {
+    return message;
+  }
+
+  return errorCode;
+};
+
+export class LogtoClientError extends Error {
+  code: LogtoClientErrorCode;
+  data: unknown;
+
+  constructor(code: LogtoClientErrorCode, data?: unknown) {
+    super(getMessageByErrorCode(code));
+    this.code = code;
+    this.data = data;
+  }
+}
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -1,30 +1,100 @@
-import LogtoClient from './index';
+import { Optional } from '@silverhand/essentials';
+
+import LogtoClient, { LogtoClientError, LogtoSignInSessionItem } from '.';
 
 const clientId = 'client_id_value';
 const endpoint = 'https://logto.dev';
 const authorizationEndpoint = `${endpoint}/oidc/auth`;
+const mockedCodeVerifier = 'code_verifier_value';
+const mockedState = 'state_value';
+const mockedSignInUri = `${authorizationEndpoint}?foo=bar`;
 const redirectUri = 'http://localhost:3000/callback';
 const requester = jest.fn();
-const signInUri = `${authorizationEndpoint}?foo=bar`;
 
-jest.mock('@logto/js', () => {
-  return {
-    ...jest.requireActual('@logto/js'),
-    fetchOidcConfig: async () => ({ authorizationEndpoint }),
-    generateSignInUri: () => signInUri,
-  };
-});
+jest.mock('@logto/js', () => ({
+  ...jest.requireActual('@logto/js'),
+  fetchOidcConfig: async () => ({ authorizationEndpoint }),
+  generateCodeChallenge: jest.fn().mockResolvedValue('code_challenge_value'),
+  generateCodeVerifier: jest.fn(() => mockedCodeVerifier),
+  generateSignInUri: jest.fn(() => mockedSignInUri),
+  generateState: jest.fn(() => mockedState),
+}));
+
+/**
+ * Make LogtoClient.signInSession accessible for test
+ */
+class LogtoClientSignInSessionAccessor extends LogtoClient {
+  public getSignInSessionItem(): Optional<LogtoSignInSessionItem> {
+    return this.signInSession;
+  }
+
+  public setSignInSessionItem(item: Optional<LogtoSignInSessionItem>) {
+    this.signInSession = item;
+  }
+}
 
 describe('LogtoClient', () => {
   test('constructor', () => {
     expect(() => new LogtoClient({ endpoint, clientId, requester })).not.toThrow();
   });
 
+  describe('signInSession', () => {
+    test('getter should throw LogtoClientError when signInSession does not contain the required property', () => {
+      const signInSessionAccessor = new LogtoClientSignInSessionAccessor({
+        endpoint,
+        clientId,
+        requester,
+      });
+
+      // @ts-expect-error
+      signInSessionAccessor.setSignInSessionItem({
+        redirectUri,
+        codeVerifier: mockedCodeVerifier,
+      });
+
+      expect(() => signInSessionAccessor.getSignInSessionItem()).toMatchError(
+        new LogtoClientError(
+          'sign_in_session.invalid',
+          new Error('At path: state -- Expected a string, but received: undefined')
+        )
+      );
+    });
+
+    test('should be able to set and get the undefined item (for clearing sign-in session)', () => {
+      const signInSessionAccessor = new LogtoClientSignInSessionAccessor({
+        endpoint,
+        clientId,
+        requester,
+      });
+
+      // @ts-expect-error
+      signInSessionAccessor.setSignInSessionItem();
+      expect(signInSessionAccessor.getSignInSessionItem()).toBeUndefined();
+    });
+
+    test('should be able to set and get the correct item', () => {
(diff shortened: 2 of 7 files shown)


simeng-li (commented):
nit: wondering whether the `signInSession` is the only session storage we gonna have. 

Just personal preference, instead of defining a `signInSession` instance within the logtoClient. We could just have to simple sessions util methods sth like

```js
getSessionStorage<T>(key: string,  guard?: SuperstructSchema): T
setSessionStorage(key:string, data: any);
```

All we need to call getSessionStorage and setSessionStorage wherever it is needed. And util methods can be tested individually.

packages/browser/src/index.ts
```diff
@@ -29,17 +33,54 @@ export type AccessToken = {
...
+    const jsonItem = sessionStorage.getItem(logtoKey);
+
+    if (!jsonItem) {
+      throw new LogtoClientError('sign_in_session.not_found');
+    }
+
+    try {
+      const item: unknown = JSON.parse(jsonItem);
+
+      return create(item, LogtoSignInSessionItemSchema);
```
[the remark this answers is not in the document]
IceHe: create can do assert and return in one line.
it's simpler.

packages/browser/src/errors.ts
```diff
@@ -0,0 +1,34 @@
...
+  const message = get(logtoClientErrorCodes, errorCode);
+
+  if (typeof message === 'string') {
+    return message;
+  }
+
+  return errorCode;
+};
+
+export class LogtoClientError extends Error {
```
gao-sun, line 25: this can be extracted into a generic (error code) class? cc @simeng-li @wangsijie @demonzoo

packages/browser/src/index.ts
```diff
@@ -29,17 +33,54 @@ export type AccessToken = {
...
+    } catch (error: unknown) {
+      if (error instanceof SyntaxError) {
```
gao-sun, line 70: Should we wrap all errors into LogtoClientError for a better dev experience?

packages/browser/src/index.ts
```diff
@@ -29,17 +33,54 @@ export type AccessToken = {
...
+        throw new LogtoClientError('sign_in_session.invalid');
+      }
+      throw error;
+    }
+  }
+
+  protected set signInSession(logtoSignInSessionItem: LogtoSignInSessionItem) {
+    assert(logtoSignInSessionItem, LogtoSignInSessionItemSchema);
```
gao-sun, line 78: We don't need assertion here since it's controlled by us

packages/browser/src/index.ts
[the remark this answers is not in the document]
IceHe: image

packages/browser/src/index.ts
```diff
@@ -29,17 +33,54 @@ export type AccessToken = {
...
   protected oidcConfig?: OidcConfigResponse;
 
+  private readonly _signInSession: Optional<LogtoSignInSessionItem>;
+
   constructor(logtoConfig: LogtoConfig) {
     this.logtoConfig = logtoConfig;
   }
 
+  protected get signInSession(): LogtoSignInSessionItem {
+    const logtoKey = getLogtoKey(this.logtoConfig.clientId);
```
gao-sun, line 58: Remember we need do it in init process for refresh / id token due to performance concern. I'm not sure if we need align this behavior over session or keep as it is. @logto-io/eng any idea?

packages/browser/src/index.ts
IceHe: I intended to prompt the JSON.parse failure due to the invalid sign-in session.

Assume that the sign-up session is all under our control, the try-catch is not necessary enough here.

Removed now. PTAL.

packages/browser/src/index.test.ts
```diff
@@ -1,30 +1,67 @@
...
+
+      const parseJsonItem = (): unknown => JSON.parse(jsonItem);
+      expect(parseJsonItem).not.toThrow();
+
+      const item: unknown = parseJsonItem();
+      expect(() => {
+        assert(item, LogtoSignInSessionItemSchema);
+      }).not.toThrow();
+
+      const logtoSignInSessionItem = item as LogtoSignInSessionItem;
```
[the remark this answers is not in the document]
IceHe: The `signInSession` getter is a protected method for now, and it's not very convenient to test.

As you said in https://github.com/logto-io/js/pull/175#pullrequestreview-881570158 , need further discussion to do it in a better way

packages/browser/src/index.test.ts
```diff
@@ -1,30 +1,96 @@
...
+  }
+}
 
 describe('LogtoClient', () => {
   test('constructor', () => {
     expect(() => new LogtoClient({ endpoint, clientId, requester })).not.toThrow();
   });
 
+  describe('signInSession', () => {
+    test('getter should tHrow LogtoClientError when signInSession has not been set', () => {
```
wangsijie, line 41: typo here.

packages/browser/src/errors.ts
```diff
@@ -0,0 +1,33 @@
...
+    not_found: 'Sign-in session not found.',
+  },
+});
+
+export type LogtoClientErrorCode = NormalizeKeyPaths<typeof logtoClientErrorCodes>;
+
+const getMessageByErrorCode = (errorCode: LogtoClientErrorCode): string => {
+  // TODO: linear issue LOG-1419
+  // eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+  const message = get(logtoClientErrorCodes, errorCode);
```
wangsijie, line 15: do we really need `lodash.get` here? seems this can work:
```ts
const message = logtoClientErrorCodes[errorCode];
```

IceHe: actual struct like this, cannot get the error message simply by one key.

image

packages/browser/src/index.ts
IceHe: Look like this? PTAL

image

packages/browser/src/errors.ts
IceHe: Added the Linear issue: [Extract generic error code class](https://linear.app/silverhand/issue/LOG-1557/extract-generic-error-code-class)

packages/browser/src/index.ts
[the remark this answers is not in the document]
IceHe: Updated. PTAL

simeng-li (approved):
overall LGTM

packages/browser/src/index.ts
```diff
@@ -29,21 +33,55 @@ export type AccessToken = {
...
+
 export default class LogtoClient {
   protected accessTokenMap = new Map<string, AccessToken>();
   protected refreshToken?: string;
   protected idToken?: string;
   protected logtoConfig: LogtoConfig;
   protected oidcConfig?: OidcConfigResponse;
+  protected logtoSignInSessionKey: string;
+
+  private readonly _signInSession: Optional<LogtoSignInSessionItem>;
```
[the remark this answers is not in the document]
IceHe: I google it and find this way to implement getter/setter in typescript.
https://www.typescripttutorial.net/typescript-tutorial/typescript-getters-setters/
https://stackoverflow.com/questions/12827266/get-and-set-in-typescript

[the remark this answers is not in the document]
IceHe: Removed. Found it unnecessary. PTAL.

packages/browser/src/index.ts
```diff
@@ -77,4 +112,10 @@ export default class LogtoClient {
 
     return this.oidcConfig;
   }
+
+  protected clearSignInSession() {
+    sessionStorage.removeItem(this.logtoSignInSessionKey);
+  }
```
gao-sun, line 118: why do we need this?

gao-sun (approved):

packages/browser/src/index.ts
IceHe: after sign-in `handleCallback`, should clear sign-in session.

IceHe: Updated. PTAL

```ts
  protected get signInSession(): Optional<LogtoSignInSessionItem> {
    const jsonItem = sessionStorage.getItem(this.logtoSignInSessionKey);

    if (!jsonItem) {
      return undefined;
    }

    try {
      const item: unknown = JSON.parse(jsonItem);
      assert(item, LogtoSignInSessionItemSchema);

      return item;
    } catch (error: unknown) {
      throw new LogtoClientError('sign_in_session.invalid', error);
    }
  }

  protected set signInSession(logtoSignInSessionItem: Optional<LogtoSignInSessionItem>) {
    if (!logtoSignInSessionItem) {
      sessionStorage.removeItem(this.logtoSignInSessionKey);

      return;
    }

    const jsonItem = JSON.stringify(logtoSignInSessionItem);
    sessionStorage.setItem(this.logtoSignInSessionKey, jsonItem);
  }
```

packages/browser/src/index.ts
IceHe: Summary on the Notion [documentation](https://www.notion.so/silverhand/TypeScript-Convention-5a8f2a6da25140318e0adbe74e6756bc#f19841d774294b0583fe1a1de1968922)

packages/browser/src/index.ts
```diff
@@ -29,21 +33,59 @@ export type AccessToken = {
...
 
   public get isAuthenticated() {
     return Boolean(this.idToken);
   }
 
+  protected get signInSession(): Optional<LogtoSignInSessionItem> {
+    const jsonItem = sessionStorage.getItem(this.logtoSignInSessionKey);
+
+    if (!jsonItem) {
+      return undefined;
```
gao-sun, line 65:
```suggestion
      return;
```

gao-sun (approved):

packages/browser/src/index.ts
IceHe: eslint does not allow.
image

PR #178, opened 2022-02-15
refactor(browser): sign out

demonzoo:
## Summary

Improve browser sign out logic

## Linear Issue Reference

[LOG-793](https://linear.app/silverhand/issue/LOG-793/browser-sign-out-method)

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -71,16 +71,20 @@ export default class LogtoClient {
   }
 
   public async signOut(postLogoutRedirectUri?: string) {
-    if (!this.idToken || !this.oidcConfig) {
+    if (!this.idToken) {
       return;
     }
 
     const { clientId, requester } = this.logtoConfig;
-    const { endSessionEndpoint, revocationEndpoint } = this.oidcConfig;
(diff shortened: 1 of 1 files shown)


packages/browser/src/index.ts
```diff
@@ -71,16 +71,16 @@ export default class LogtoClient {
...
     }
 
     const { clientId, requester } = this.logtoConfig;
-    const { endSessionEndpoint, revocationEndpoint } = this.oidcConfig;
+    const { endSessionEndpoint, revocationEndpoint } = await this.getOidcConfig();
     const logtoKey = getLogtoKey(clientId);
 
     if (this.refreshToken) {
-      void revoke(revocationEndpoint, clientId, this.refreshToken, requester);
+      await revoke(revocationEndpoint, clientId, this.refreshToken, requester);
```
gao-sun, line 83: note this step is optional, i.e. we should not block sign out process even if revoke failed

demonzoo: OK, added a try-catch block as we discussed, and do nothing in the catch block at this point, because we don't want to interrupt the sign out flow even if the token revocation is failed.

gao-sun (approved):

PR #182, opened 2022-02-15
feat(browser): browser end user information getters

demonzoo:
## Summary

Provide getter functions for browser SDK
* isAuthenticated
* getAccessToken
* getIdTokenClaims
* fetchUserInfo

## Testing

Build succeeded. Test cases will be added in a follow-up PR.

### DIFF
--- a/packages/browser/src/errors.ts
+++ b/packages/browser/src/errors.ts
@@ -5,6 +5,10 @@ const logtoClientErrorCodes = Object.freeze({
   sign_in_session: {
     invalid: 'Invalid sign-in session.',
   },
+  not_authenticated: 'Not authenticated.',
+  get_access_token_by_refresh_token_failed: 'Failed to get access token by refresh token.',
+  fetch_user_info_failed: 'Unable to fetch user info. The access token may be invalid.',
+  invalid_id_token: 'Invalid id token.',
 });
 
 export type LogtoClientErrorCode = NormalizeKeyPaths<typeof logtoClientErrorCodes>;
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -1,21 +1,28 @@
 import {
   createRequester,
+  decodeIdToken,
   fetchOidcConfig,
+  fetchTokenByRefreshToken,
+  fetchUserInfo as fetchUserInfoCoreFunction,
   generateCodeChallenge,
   generateCodeVerifier,
   generateSignInUri,
   generateSignOutUri,
   generateState,
+  IdTokenClaims,
   OidcConfigResponse,
   Requester,
   revoke,
+  UserInfoResponse,
+  verifyIdToken as verifyIdTokenCoreFunction,
   withReservedScopes,
 } from '@logto/js';
 import { Nullable, Optional } from '@silverhand/essentials';
+import { createRemoteJWKSet } from 'jose';
 import { assert, Infer, string, type } from 'superstruct';
 
 import { LogtoClientError } from './errors';
-import { getDiscoveryEndpoint, getLogtoKey } from './utils';
+import { buildAccessTokenKey, getDiscoveryEndpoint, getLogtoKey } from './utils';
 
 export * from './errors';
 
@@ -90,6 +97,75 @@ export default class LogtoClient {
     sessionStorage.setItem(this.logtoStorageKey, jsonItem);
   }
 
+  public async getAccessToken(resource?: string): Promise<Optional<string>> {
+    if (!this.idToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    const accessTokenKey = buildAccessTokenKey(resource);
+    const accessToken = this.accessTokenMap.get(accessTokenKey);
+
+    if (accessToken && accessToken.expiresAt > Date.now()) {
+      return accessToken.token;
+    }
+
+    // Token expired, remove it from the map
+    if (accessToken) {
+      this.accessTokenMap.delete(accessTokenKey);
+    }
+
+    // Fetch new access token by refresh token
+    const { clientId } = this.logtoConfig;
+
+    if (!this.refreshToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    try {
+      const { tokenEndpoint } = await this.getOidcConfig();
+      const { accessToken, refreshToken, idToken, scope, expiresIn } =
+        await fetchTokenByRefreshToken(
+          { clientId, tokenEndpoint, refreshToken: this.refreshToken, resource },
+          this.requester
+        );
+      this.accessTokenMap.set(accessTokenKey, {
+        token: accessToken,
+        scope,
+        expiresAt: Math.round(Date.now() / 1000) + expiresIn,
+      });
+
+      localStorage.setItem(`${this.logtoStorageKey}:refreshToken`, refreshToken);
+
+      if (idToken) {
+        await this.verifyIdToken(idToken);
+        localStorage.setItem(`${this.logtoStorageKey}:idToken`, idToken);
+      }
+
+      return accessToken;
+    } catch (error: unknown) {
+      throw new LogtoClientError('get_access_token_by_refresh_token_failed', error);
+    }
+  }
+
+  public getIdTokenClaims(): IdTokenClaims {
+    if (!this.idToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    return decodeIdToken(this.idToken);
+  }
+
+  public async fetchUserInfo(): Promise<UserInfoResponse> {
+    const { authorizationEndpoint } = await this.getOidcConfig();
+    const accessToken = await this.getAccessToken();
+
+    if (!accessToken) {
+      throw new LogtoClientError('fetch_user_info_failed');
+    }
+
+    return fetchUserInfoCoreFunction(authorizationEndpoint, accessToken, this.requester);
+  }
+
   public async signIn(redirectUri: string) {
     const { clientId, resources, scopes: customScopes } = this.logtoConfig;
     const { authorizationEndpoint } = await this.getOidcConfig();
@@ -114,7 +190,7 @@ export default class LogtoClient {
 
   public async signOut(postLogoutRedirectUri?: string) {
     if (!this.idToken) {
-      return;
+      throw new LogtoClientError('not_authenticated');
     }
 
     const { clientId } = this.logtoConfig;
@@ -144,7 +220,7 @@ export default class LogtoClient {
     window.location.assign(url);
   }
 
-  protected async getOidcConfig(): Promise<OidcConfigResponse> {
+  private async getOidcConfig(): Promise<OidcConfigResponse> {
     if (!this.oidcConfig) {
       const { endpoint } = this.logtoConfig;
       const discoveryEndpoint = getDiscoveryEndpoint(endpoint);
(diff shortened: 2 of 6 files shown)


packages/browser/src/index.ts
```diff
@@ -86,6 +86,47 @@ export default class LogtoClient {
...
+    const accessToken = this.accessTokenMap.get(accessTokenKey);
+
+    if (!accessToken) {
+      throw new LogtoClientError('access_token.failed');
+    }
+
+    if (accessToken.expiresAt < Date.now()) {
+      this.accessTokenMap.delete(accessTokenKey);
+
+      throw new LogtoClientError('access_token.expired');
```
gao-sun, line 104: refresh logic missing?

packages/browser/src/index.ts
```diff
@@ -86,6 +86,47 @@ export default class LogtoClient {
     sessionStorage.setItem(this.logtoSignInSessionKey, jsonItem);
   }
 
+  public isAuthenticated() {
+    return Boolean(this.idToken);
+  }
+
+  public getAccessToken(resource = '', scopes = []): Optional<string> {
```
gao-sun, line 93: per discussed today `scopes` should be removed

packages/browser/src/index.ts
```diff
@@ -86,6 +86,47 @@ export default class LogtoClient {
     sessionStorage.setItem(this.logtoSignInSessionKey, jsonItem);
   }
 
+  public isAuthenticated() {
```
gao-sun, line 89: better to use property getter here?

packages/browser/src/index.ts
```diff
@@ -86,6 +86,47 @@ export default class LogtoClient {
...
+    if (accessToken.expiresAt < Date.now()) {
+      this.accessTokenMap.delete(accessTokenKey);
+
+      throw new LogtoClientError('access_token.expired');
+    }
+
+    return accessToken.token;
+  }
+
+  public getIdTokenClaims(): Optional<IdTokenClaims> {
```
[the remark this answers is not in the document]
demonzoo: When idToken is empty, the function will throw an error, thus `optional`

packages/browser/src/index.ts
gao-sun: also we allow empty / null but no default value for these params

packages/browser/src/index.ts
demonzoo: I think both are fine, but I don't see one better than the other. And since the other three getter functions all have parameters, they are not allowed to use property getter pattern.
Therefore, for the sake of consistency, we should all define it in the normal function pattern?

packages/browser/src/index.ts
demonzoo: Good call, thanks. I completely dropped that part. Code updated just now, please review again.

packages/browser/src/index.ts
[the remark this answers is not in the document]
gao-sun: actually the point of throwing an error is we can assert non-optional?

packages/browser/src/index.ts
gao-sun: from the naming it's like a prop actually, unless it's called `getIsAuthenticated` or something

packages/browser/src/index.ts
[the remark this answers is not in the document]
demonzoo: Oh, I think you are right. 😂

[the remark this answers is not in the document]
demonzoo: Updated

packages/browser/src/index.ts
demonzoo: Updated, thanks

packages/browser/src/index.ts
demonzoo: Thanks, code updated

packages/browser/src/index.ts
```diff
@@ -88,6 +93,75 @@ export default class LogtoClient {
...
+  public async getAccessToken(resource?: string): Promise<Optional<string>> {
+    if (!this.idToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    const accessTokenKey = buildAccessTokenKey(resource);
+    const accessToken = this.accessTokenMap.get(accessTokenKey);
+
+    if (!accessToken) {
+      throw new LogtoClientError('get_access_token_failed');
```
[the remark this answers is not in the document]
simeng-li: i.e. if the user wants to request a new resource using a refreshToken

packages/browser/src/index.ts
```diff
@@ -88,6 +93,75 @@ export default class LogtoClient {
...
+    }
+
+    if (accessToken.expiresAt < Date.now()) {
+      this.accessTokenMap.delete(accessTokenKey);
+
+      const { clientId, requester } = this.logtoConfig;
+      const { tokenEndpoint } = await this.getOidcConfig();
```
simeng-li, line 112: just want to double-check will getOidcConfig throws if no proper config is fetched?

packages/browser/src/index.ts
```diff
@@ -88,6 +93,75 @@ export default class LogtoClient {
...
+        this.accessTokenMap.set(accessTokenKey, {
+          token: accessToken,
+          scope,
+          expiresAt: Math.round(Date.now() / 1000) + expiresIn,
+        });
+
+        localStorage.setItem(`${this.logtoStorageKey}:refreshToken`, refreshToken);
+
+        if (idToken) {
+          localStorage.setItem(`${this.logtoStorageKey}:idToken`, idToken);
```
simeng-li, line 133: @xiaoyijun do we need to validate the `idToken` here?

simeng-li (commented):
also d'like to check whether it is possible to add some test cases along with this PR or defer to the future.

packages/browser/src/errors.ts
```diff
@@ -5,6 +5,10 @@ const logtoClientErrorCodes = Object.freeze({
   sign_in_session: {
     invalid: 'Invalid sign-in session.',
   },
+  not_authenticated: 'Not authenticated.',
+  get_access_token_failed: 'Failed to get access token.',
+  get_access_token_by_refresh_token_failed: 'Failed to get access token by refresh token.',
+  fetch_user_info_failed: 'Unable to fetch user info. Getting access token failed.',
```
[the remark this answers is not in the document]
demonzoo: This error is thrown when the access token is not available, thus failed to get user info. So I thought mentioning the root cause might be appreciated.

packages/browser/src/index.ts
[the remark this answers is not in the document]
demonzoo: Good call! Code updated, thanks

packages/browser/src/index.ts
demonzoo: Yes, it will throw errors

packages/browser/src/index.ts
```diff
@@ -88,6 +93,73 @@ export default class LogtoClient {
...
 
+  public async getAccessToken(resource?: string): Promise<Optional<string>> {
+    if (!this.idToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    const accessTokenKey = buildAccessTokenKey(resource);
+    const accessToken = this.accessTokenMap.get(accessTokenKey);
+
+    if (!accessToken || accessToken.expiresAt < Date.now()) {
```
gao-sun, line 104: Why don't we just return access token here if it's still valid? Then we can avoid 套娃 here

packages/browser/src/index.ts
gao-sun: Then we can just remove try catch in this function if we're going to naturally throw these errors. Or we wrap all of them into one error scope. Say `get_access_token.unable_fetch_oidc_config`

packages/browser/src/index.ts
demonzoo: Yes we do need to verify idToken here, will update shortly

packages/browser/src/index.ts
```diff
@@ -88,6 +93,75 @@ export default class LogtoClient {
...
 
+  public async getAccessToken(resource?: string): Promise<Optional<string>> {
+    if (!this.idToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    const accessTokenKey = buildAccessTokenKey(resource);
+    const accessToken = this.accessTokenMap.get(accessTokenKey);
+
+    if (!accessToken) {
```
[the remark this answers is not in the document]
demonzoo: Yes, code has been updated to try to get access token by exchanging refresh token, if no access token is presented in the first place.

packages/browser/src/index.ts
demonzoo: Sorry, I don't quite get your points here. The valid access token is returned at the end of this function.
The `if` block here is the condition branch for those empty / invalid access tokens, which will not be executed in the `valid token` situation

packages/browser/src/index.ts
```diff
@@ -88,6 +93,75 @@ export default class LogtoClient {
     sessionStorage.setItem(this.logtoStorageKey, jsonItem);
   }
 
+  public async getAccessToken(resource?: string): Promise<Optional<string>> {
+    if (!this.idToken) {
+      throw new LogtoClientError('not_authenticated');
+    }
+
+    const accessTokenKey = buildAccessTokenKey(resource);
+    const accessToken = this.accessTokenMap.get(accessTokenKey);
```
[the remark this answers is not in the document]
demonzoo: Discussed offline and resolving as "not possible"

packages/browser/src/index.ts
demonzoo: Yeah, I see your points. After checking the code, I found `tokenEndpoint` is only needed in the try-catch block, so I moved it to the `try` block and if `getOidcConfig` is failed, the error will be wrapped into the same error scope: `get_access_token_by_refresh_token_failed`.

Does this sound good to you? @gao-sun @simeng-li

packages/browser/src/errors.ts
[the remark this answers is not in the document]
simeng-li: Yeah, I mean could it be access token invalid instead of. get access token failed?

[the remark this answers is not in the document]
demonzoo: ok, np

packages/browser/src/index.ts
simeng-li:
```
if (accessToken && accessToken.expiresAt >= Date.now()) return accessToken:
if (accessToken) {  this.accessTokenMap.delete(accessTokenKey); }
...
```

demonzoo: Ah, I see

simeng-li: avoid nested if conditions, just return early, and flatten all the logics if possible.

demonzoo: Thanks, code updated

packages/browser/src/errors.ts
```diff
@@ -5,6 +5,10 @@ const logtoClientErrorCodes = Object.freeze({
   sign_in_session: {
     invalid: 'Invalid sign-in session.',
   },
+  not_authenticated: 'Not authenticated.',
+  get_access_token_by_refresh_token_failed: 'Failed to get access token by refresh token.',
+  fetch_user_info_failed: 'Unable to fetch user info. Invalid access token.',
```
gao-sun, line 10:
```suggestion
  fetch_user_info_failed: 'Unable to fetch user info. The access token may be invalid.',
```

packages/browser/src/errors.ts
```diff
@@ -5,6 +5,10 @@ const logtoClientErrorCodes = Object.freeze({
...
+  invalid_id_token: 'Invalid id token.',
```
gao-sun, line 11: it'll be better to use scoped keys here, say `{ get_access_token: { not_authenticated: 'Not ...', ... } }`

gao-sun (approved):

packages/browser/src/errors.ts
demonzoo: Updated, thanks

PR #183, opened 2022-02-16
refactor(browser): read tokens from local storage on init

demonzoo:
## Summary

Read `idToken` and `refreshToken` from local storage and set it to `LogtoClient` instance on initialize.

## Testing

Passed all test cases

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -10,7 +10,7 @@ import {
   revoke,
   withReservedScopes,
 } from '@logto/js';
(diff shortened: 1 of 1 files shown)


xiaoyijun (approved):
LGTM

xiaoyijun:
We need to update the storage once we have got a new token response. (e.g. get token by refresh token)

demonzoo:
> We need to update the storage once we have got a new token response. (e.g. get token by refresh token)

Sure we do. That part will be added in a follow-up PR shortly. Thanks!

PR #184, opened 2022-02-16
test(browser): add browser sign out test cases

demonzoo:
## Summary

Add unit test cases for sign out in browser SDK

## Linear Issue Reference

[LOG-1571](https://linear.app/silverhand/issue/LOG-1571/add-unit-test-cases-for-browser-sign-out)

## Testing

Passed all test cases

### DIFF
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -5,15 +5,21 @@ import LogtoClient, { LogtoClientError, LogtoSignInSessionItem } from '.';
 const clientId = 'client_id_value';
 const endpoint = 'https://logto.dev';
 const authorizationEndpoint = `${endpoint}/oidc/auth`;
+const endSessionEndpoint = `${endpoint}/oidc/session/end`;
+const revocationEndpoint = `${endpoint}/oidc/token/revocation`;
 const mockedCodeVerifier = 'code_verifier_value';
 const mockedState = 'state_value';
 const mockedSignInUri = `${authorizationEndpoint}?foo=bar`;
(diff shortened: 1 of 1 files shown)


packages/browser/src/index.test.ts
```diff
@@ -97,4 +102,36 @@ describe('LogtoClient', () => {
...
+
+  describe('signOut', () => {
+    beforeEach(() => {
+      localStorage.setItem(idTokenStorageKey, 'id_token_value');
+      localStorage.setItem(refreshTokenStorageKey, 'refresh_token_value');
+    });
+
+    test('should call token revocation endpoint with requester', async () => {
+      const logtoClient = new LogtoClient({ endpoint, clientId, requester });
+      await logtoClient.signOut(postSignOutRedirectUri);
```
gao-sun, line 114: will this send a real request?

gao-sun: also should cover request failed?

demonzoo: > will this send a real request?

No, the `requester` function is mocked at the very beginning of this test suite.

> also should cover request failed?

I'm afraid not, as we do not want to block sign-out flow even if the revocation is failed, so there is zero handling in the `try-catch` block of the `revoke` request.

gao-sun: yes, i mean we should cover "even if the request failed but sign out is still non-blocking" case

demonzoo: OK, will add that part shortly

gao-sun (approved):

PR #185, opened 2022-02-16
refactor(js): decodeIdToken

IceHe:
## Summary

refactor(js): decodeIdToken

- replace `superstruct.create` with `superstruct.assert`

---

Since `create` is different from `assert` in `superstruct`
- [assert](https://docs.superstructjs.org/api-reference/core#assert): just validate
- [create](https://docs.superstructjs.org/api-reference/core#create): validate, fill with default values and etc.
    - with [coercions](https://docs.superstructjs.org/api-reference/coercions): defaulted and trimmed

## Linear Issue Reference

LOG-1355

## Testing

Pass all existing tests.

---
@logto-io/eng


xiaoyijun (approved):
LGTM

PR #165, opened 2022-02-08
feat(js): fetch access token by authorization code

demonzoo:
## Summary

Fetch access token by authorization code

## Testing

- [x] Passed all unit tests

PR #186, opened 2022-02-16
feat(js): add CodeTokenResponse

IceHe:
## Summary

feat(js): add CodeTokenResponse ( which is missing in #165 )

---

The core difference between `CodeTokenResponse` and `RefreshTokenTokenResponse` is `idToken`.

- `CodeTokenResponse.idToken` is required.
- `RefreshTokenTokenResponse.idToken` is optional.

image

## Linear Issue Reference

LOG-770

## Testing

Pass all existing tests.

---
@logto-io/eng

### DIFF
--- a/packages/js/src/core/fetch-token.ts
+++ b/packages/js/src/core/fetch-token.ts
@@ -21,7 +21,17 @@ export type FetchTokenByRefreshTokenParameters = {
   scopes?: string[];
 };
 
-type TokenSnakeCaseResponse = {
+type SnakeCaseCodeTokenResponse = {
+  access_token: string;
+  refresh_token: string;
+  id_token: string;
+  scope: string;
+  expires_in: number;
+};
+
+export type CodeTokenResponse = KeysToCamelCase<SnakeCaseCodeTokenResponse>;
+
+type SnakeCaseRefreshTokenTokenResponse = {
   access_token: string;
   refresh_token: string;
   id_token?: string;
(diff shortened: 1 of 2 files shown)


packages/js/src/core/fetch-token.ts
```diff
@@ -21,15 +22,21 @@ export type FetchTokenByRefreshTokenParameters = {
...
-  id_token?: string;
+  id_token: string;
   scope: string;
   expires_in: number;
 };
 
-export type RefreshTokenTokenResponse = KeysToCamelCase<TokenSnakeCaseResponse>;
+type SnakeCaseRefreshTokenTokenResponse = Except<SnakeCaseCodeTokenResponse, 'id_token'> & {
+  id_token?: string;
+};
```
xiaoyijun, line 35: A new type `RefreshTokenTokenResponse` would be better?
This line of code seems hacked.

xiaoyijun (changes requested):

packages/js/src/core/fetch-token.ts
xiaoyijun: `RefreshTokenTokenResponse` will change in the future.

IceHe: I used to define two types separately.
And i intended to point out the core difference between `SnakeCaseCodeTokenResponse` and `SnakeCaseRefreshTokenTokenResponse` here.

I'm good with both.

IceHe: Define two types separately. PTAL

xiaoyijun (approved):
LGTM

PR #187, opened 2022-02-16
refactor(browser): make requester an optional LogtoClient constructor parameter

IceHe:
## Summary

refactor(browser): make requester an optional LogtoClient constructor parameter

Make LogtoClient easier to use: the developer does not need to create Requester his/herself.

## Linear Issue Reference

LOG-1575

## Testing

Pass all existing tests.

---
@logto-io/eng

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -1,4 +1,5 @@
 import {
+  createRequester,
   fetchOidcConfig,
   generateCodeChallenge,
   generateCodeVerifier,
@@ -24,7 +25,6 @@ export type LogtoConfig = {
   scopes?: string[];
   resources?: string[];
   usingPersistStorage?: boolean;
-  requester: Requester;
 };
 
 export type AccessToken = {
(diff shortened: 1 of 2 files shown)


xiaoyijun (approved):
LGTM

packages/browser/src/index.ts
```diff
@@ -48,12 +48,14 @@ export default class LogtoClient {
...
+  protected requester: Requester;
 
-  constructor(logtoConfig: LogtoConfig) {
+  constructor(logtoConfig: LogtoConfig, requester = createRequester()) {
     this.logtoConfig = logtoConfig;
     this.logtoStorageKey = getLogtoKey(logtoConfig.clientId);
-    this.idToken = conditional(localStorage.getItem(`${this.logtoStorageKey}:idToken`));
+    this.requester = requester;
     this.refreshToken = conditional(localStorage.getItem(`${this.logtoStorageKey}:refreshToken`));
+    this.idToken = conditional(localStorage.getItem(`${this.logtoStorageKey}:idToken`));
```
gao-sun, line 58: should be nullable instead of undefined?

gao-sun (approved):

packages/browser/src/index.ts
IceHe: just intended to sort the statements in the same order of the related object properties here.

```ts
  ……
  protected refreshToken?: string;
  protected idToken?: string;
```

we can do nullable later.
LOG-1576

PR #195, opened 2022-02-22
feat(browser-sample): init package

IceHe:
## Summary

- feat(browser-sample): init package

## Linear Issue Reference

LOG-1595

## Testing

Build succeeded

image


packages/browser-sample/package.json
```diff
@@ -0,0 +1,45 @@
+{
+  "name": "browser-sample",
+  "version": "0.1.0",
+  "license": "UNLICENSED",
```
gao-sun, line 4: MIT will be good

packages/playground/src/consts/url.ts
```diff
@@ -1,2 +1,3 @@
 export const baseUrl = `http://${String(process.env.HOST)}:${String(process.env.PORT)}`;
-export const oidcDomain = process.env.RAZZLE_OIDC_BASE_DOMAIN ?? 'N/A';
+// eslint-disable-next-line @typescript-eslint/no-unsafe-assignment
+export const oidcDomain: string = process.env.RAZZLE_OIDC_BASE_DOMAIN ?? 'N/A';
```
gao-sun, line 3: why is this unsafe?

demonzoo: Because we don't have type defined for env variables at this point. So the type of `process.env.RAZZLE_OIDC_BASE_DOMAIN` is `any`, and you cannot assign `any` to another variable, thus the `unsafe-assignment`.
We will have to either use `.env` files or include `@types/node` in order to infer env variable types, I think.

packages/browser-sample/package.json
demonzoo: Updated to use MIT license

packages/playground/src/consts/url.ts
demonzoo: I somehow find it easier to solve this lint issue by simply wrapping it inside a `String()`... Please review

demonzoo (approved):

PR #199, opened 2022-02-24
fix(js): response_type should be hard-coded as 'code'

demonzoo:
## Summary

Fix incorrect `response_type` used in js SDK. Should use `code` instead of `authorization_code`

## Testing

Passed all test cases.


xiaoyijun (approved):
LGTM

PR #200, opened 2022-02-24
fix(js): grant type should be 'authorization_code'  instead of 'code'

demonzoo:
## Summary

Fix incorrect `grant_type` usage in js SDK. The value of enum `GrantType.Authorization` should be `authorization_code` instead of `code`

## Testing

Passed all test cases


xiaoyijun (approved):
LGTM

PR #206, opened 2022-02-28
ci(codecov): update codecov job

simeng-li:
## Summary

update codecov job

deprecate codecov node uploader

## Linear Issue Reference

LOG-1276

## Testing
@logto-io/eng


codecov.yml
```diff
@@ -1,5 +1,22 @@
...
+        flags:
+          - browser
+
+flags:
+  js:
+    paths:
+      - ./packages/js
+  browser:
+    paths:
+      - ./packages/browser
```
gao-sun, line 22: by design?

simeng-li: good call, should not need that, as we have a general auto rule set. Let me check

wangsijie (approved):

PR #211, opened 2022-03-08
fix(browser): should use userinfo_endpoint in fetchUserInfo function

demonzoo:
## Summary

`fetchUserInfo` function should use `userinfo_endpoint` instead of `authorization_endpoint`. This PR addresses the issue.

## Linear Issue Reference

[LOG-1853](https://linear.app/silverhand/issue/LOG-1853/fix-wrong-endpoint-used-in-fetchuserinfo-in-browser-sdk)

## Testing

- [x] passed unit tests

### DIFF
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -7,6 +7,7 @@ const clientId = 'client_id_value';
 const endpoint = 'https://logto.dev';
 
 const authorizationEndpoint = `${endpoint}/oidc/auth`;
+const userinfoEndpoint = `${endpoint}/oidc/me`;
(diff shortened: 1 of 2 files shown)


gao-sun (commented):
why no unit test is edited? should we add some assertion for this?

demonzoo:
> why no unit test is edited? should we add some assertion for this?

I'm currently working on it. Previously, the unit test code doesn't cover the URI that each functions calls. I'm thinking about using `jest.toBeCalledWith` to add more assertions.

demonzoo:
Unit test updated @gao-sun 
Will create another PR to improve other test cases after this is merged

gao-sun (approved):
cool

PR #212, opened 2022-03-08
feat(react-sample): add react sample webpages

demonzoo:
## Summary

Add react sample webpages

## Testing

- [x] Authentication flow works

### DIFF
--- a/packages/react-sample/package.json
+++ b/packages/react-sample/package.json
@@ -5,15 +5,17 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "precommit": "lint-staged",
-    "start": "parcel src/index.html",
+    "start": "parcel src/index.html -p 3000",
     "check": "tsc --noEmit",
     "build": "pnpm check && rm -rf dist && parcel build src/index.html --no-autoinstall",
     "lint": "eslint --ext .ts --ext .tsx src",
     "stylelint": "stylelint \"src/**/*.scss\""
   },
   "dependencies": {
+    "@logto/react": "^0.1.0",
     "react": "^17.0.2",
-    "react-dom": "^17.0.2"
+    "react-dom": "^17.0.2",
+    "react-router-dom": "^6.2.2"
   },
   "devDependencies": {
     "@parcel/core": "^2.3.2",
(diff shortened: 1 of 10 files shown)


packages/react-sample/package.json
```diff
@@ -5,15 +5,17 @@
   "scripts": {
     "preinstall": "npx only-allow pnpm",
     "precommit": "lint-staged",
-    "start": "parcel src/index.html",
+    "start": "parcel src/index.html -p 3000",
```
gao-sun, line 8: will it be better to align parcel port usage across the reops? `PORT=3000 parcel src/index.html`

packages/react-sample/src/pages/Callback/index.tsx
```diff
@@ -0,0 +1,18 @@
...
+
+  useEffect(() => {
+    if (isAuthenticated && !isLoading) {
+      navigate('/', { replace: true });
+    }
+  }, [isAuthenticated, isLoading, navigate]);
+
+  return <p>Redirecting...</p>;
+};
+
```
gao-sun, line 17: should have `handleSignInCallback()` logic?

demonzoo: Not necessary since there is a `useEffect` in the react SDK that automatically calls `handleSignInCallback`.
```typescript
  useEffect(() => {
    if (!isAuthenticated && logtoClient?.isSignInRedirected(window.location.href)) {
      void handleSignInCallback(window.location.href);
    }
  }, [handleSignInCallback, isAuthenticated, logtoClient]);
```

```typescript
  public isSignInRedirected(url: string): boolean {
    const { signInSession } = this;

    if (!signInSession) {
      return false;
    }
    const { redirectUri } = signInSession;

    return url.startsWith(redirectUri);
  }
```

packages/react-sample/package.json
demonzoo: I'm OK with either way. But I just noticed we have other parcel options specified in `logto/console`
```
"dev": "PORT=5002 parcel src/index.html --public-url /console --no-hmr --no-cache",
```
Would it be better using `-p` or `--port` to align with other option usages?

wangsijie (approved):

packages/react-sample/package.json
demonzoo: ok, after a quick heads-up, we decided to use `-p` across all repos. Will create another PR to address this

packages/react-sample/src/pages/Callback/index.tsx
demonzoo: Discussed offline and this is OK.

PR #214, opened 2022-03-09
refactor(browser): use strict check logic in isSignInRedirected function

demonzoo:
## Summary

Improve `isSignInRedirected` logic in browser SDK. Instead of using `startWith`, we should use strict check on the redirect URL, the origin and pathname must be exactly matched. 

## Linear Issue Reference

[LOG-1860](https://linear.app/silverhand/issue/LOG-1860/improve-issigninredirected-logic-in-browser-sdk)

## Testing

- [x] passed unit tests

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -243,8 +243,9 @@ export default class LogtoClient {
       return false;
     }
     const { redirectUri } = signInSession;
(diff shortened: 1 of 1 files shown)


gao-sun (approved):

IceHe (approved):
LGTM

PR #220, opened 2022-03-10
ci: upgrade node workflow

gao-sun:
## Summary

upgrade node workflow version to use node 16 as default

## Testing

CI


IceHe (approved):
LGTM

PR #232, opened 2022-03-16
refactor(browser): delete expired access token in getAccessToken

IceHe:
## Summary

- refactor(browser): delete expired access token in getAccessToken
- test(browser): delete expired access token in getAccessToken

Related PR #229 [comment](https://github.com/logto-io/js/pull/229#discussion_r826959443)

## Linear Issue Reference

LOG-1909

## Testing

image

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -159,8 +159,13 @@ export default class LogtoClient {
       return accessToken.token;
     }
 
+    // Since the access token has expired, delete it from the map.
+    if (accessToken) {
+      this.accessTokenMap.delete(accessTokenKey);
+    }
+
     /**
-     * Token has expired, need to fetch a new one using refresh token.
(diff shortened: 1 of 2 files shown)


packages/browser/src/index.ts
```diff
@@ -159,8 +159,13 @@ export default class LogtoClient {
       return accessToken.token;
     }
 
+    // Since the access token has expired, delete it from the map.
+    if (accessToken) {
+      this.accessTokenMap.delete(accessTokenKey);
```
gao-sun, line 164: per our discussion in the previous PR, this will cause duplicate removal.

IceHe: Tested `expect(accessTokenMap.delete).toBeCalledTimes(1);` above.
Seem that it will not cause duplicate removal.

Maybe the test case above is false.

demonzoo (approved):

PR #236, opened 2022-04-21
refactor(js): rename clientId to appId in js SDKs

demonzoo:
## Summary

Rename `clientId` to `appId` in JS SDKs

## Linear Issue Reference

[LOG-2252](https://linear.app/silverhand/issue/LOG-2252/rename-client-id-to-app-id-in-sdks)

## Testing

- [x] Passed all unit tests
- [x] Tested browser sample and it worked
- [x] Tested react sample and it also worked

### DIFF
--- a/packages/browser-sample/src/index.js
+++ b/packages/browser-sample/src/index.js
@@ -1,13 +1,12 @@
 import LogtoClient from '@logto/browser';
 
-import { endpoint, clientId } from './consts.js';
+import { endpoint, appId } from './consts.js';
 import Callback from './pages/Callback.js';
 import Home from './pages/Home.js';
 
 import './index.scss';
 
-const logtoClient = new LogtoClient({ endpoint, clientId });
-
+const logtoClient = new LogtoClient({ endpoint, appId });
 const app = document.querySelector('#app');
 
 // Could replace this with a formal router solution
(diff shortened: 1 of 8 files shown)


packages/browser-sample/src/index.js
```diff
@@ -1,12 +1,13 @@
...
-import { endpoint, clientId } from './consts.js';
+import { endpoint, appId } from './consts.js';
 import Callback from './pages/Callback.js';
 import Home from './pages/Home.js';
 
 import './index.scss';
 
-const logtoClient = new LogtoClient({ endpoint, clientId });
+const logtoClient = new LogtoClient({ endpoint, appId });
+console.log('logto client:', logtoClient);
```
wangsijie, line 10: Is this on purpose?

wangsijie (approved):
Overall LGTM.

packages/browser-sample/src/index.js
demonzoo: Oops, should remove it. Thanks

packages/js/src/core/fetch-token.test.ts
```diff
@@ -61,7 +61,7 @@ describe('fetch access token by providing valid refresh token', () => {
 
     const tokenResponse = await fetchTokenByRefreshToken(
       {
-        clientId: 'client_id_value',
```
gao-sun, line 64: we can keep core client id since it directly contacts oidc?

demonzoo: OK, reverted all the changes made in js core, keeping only those in `browser` and above layers

PR #253, opened 2022-04-28
fix(browser): read/write refreshToken from/to localStorage only

demonzoo:
## Summary

Read/write refreshToken from localStorage only, instead of loading from browser memory object.

This fix aims to prevent using an expired refreshToken when more application tabs are open. Since we have the refreshToken rotations enabled, refreshToken will expire immediately after it is used to exchange for the accessToken.

## Linear Issue Reference

[LOG-2328](https://linear.app/silverhand/issue/LOG-2328/always-getset-refresh-token-from-local-storage)

## Testing

- [x] Tested react sample project in two browser tabs, both worked fine.

### DIFF
--- a/packages/browser/src/index.ts
+++ b/packages/browser/src/index.ts
@@ -69,14 +69,12 @@ export default class LogtoClient {
   protected readonly accessTokenMap = new Map<string, AccessToken>();
 
   private readonly getAccessTokenPromiseMap = new Map<string, Promise<string>>();
-  private _refreshToken: Nullable<string>;
   private _idToken: Nullable<string>;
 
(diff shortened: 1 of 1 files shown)


simeng-li (approved):
LGTM:  BTW wondering is it possible to have some tests cover this issue?  always get the refresh token from storage

demonzoo:
> LGTM: BTW wondering is it possible to have some tests cover this issue? always get the refresh token from storage

I thought about it but since this addresses an issue that occurrs when using logto across multiple browser tabs, I didn't think there was a way to cover this case in unit test.
Let me know if you have any ideas, thanks

IceHe (approved):
LGTM

PR #270, opened 2022-05-17
fix(browser): clear authenticated status before signing-in

demonzoo:
## Summary

Remove `idToken` and `refreshToken` from localStorage before navigating to sign-in page, in order to clear the authenticated status in application.

This is helpful when session is expired and user would need to re-sign-in to the app. Restoring these two token would help provider a cleaner environment in the re-signing-in process.

## Testing

- [x] Passed UT
- [x] React sample works


IceHe (approved):
LGTM

PR #274, opened 2022-05-19
fix(browser-sample): add typescript to browser-sample dev dependencies

demonzoo:
## Summary

As discussed offline with @gao-sun , add typescript to devDependencies of browser-sample project, in order to get rid of the annoying warnings from eslint-configs.

## Testing

- [x] `pnpm i` does not print error messages of missing `typescript` peer dependencies


IceHe (approved):
LGTM

PR #284, opened 2022-05-27
fix: should export 'LogtoClientError' as a class instead of a type

demonzoo:
## Summary

Fix an issue in React and Vue SDKs.

The exported `LogtoClientError` should be a class not a type.

## Testing

- [x] Tested locally in admin console, should work with expressions like `error instanceof LogtoClientError`


IceHe (approved):
LGTM

PR #285, opened 2022-05-27
refactor: expose OidcError class from SDKs

demonzoo:
## Summary

Previously, in our LogtoError class, there is a `data` property and its type is `unknown`, which makes the it difficult for any client that integrated our SDK to load content in `error.data`. The user will have to write a lousy type guard in order to make it work.

Therefore, we provide a new `OidcError` class from our core SDK, and export it in all js SDKs, in order to make it possible to write:

```ts
if (error.data instanceof OidcError) {
  const { error, errorDesccription } = error.data;
  ...
}
```


IceHe (approved):
LGTM

PR #292, opened 2022-05-31
fix(react): calling getIdTokenClaims will not cause infinite loop

charIeszhao:
## Summary

Remove `handleError` from `getIdTokenClaims` catch block, in order to avoid an infinite loop issue when calling it on component is first rendered.

## Linear Issue Reference

Related internal discusstion:
[LOG-2799](https://linear.app/silverhand/issue/LOG-2799/discussion-infinite-loop-on-calling-getidtokenclaims-on-component)

## Testing

- [x] Tested in react sample project, calling `getIdTokenClaims` outside `useEffect` didn't end up in infinite loop any more.

### DIFF
--- a/packages/react/src/hooks/index.ts
+++ b/packages/react/src/hooks/index.ts
@@ -178,10 +178,10 @@ const useLogto = (): Logto => {
 
     try {
       return logtoClient.getIdTokenClaims();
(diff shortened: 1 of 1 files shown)


xiaoyijun (approved):
LGTM

PR #293, opened 2022-05-31
refactor: remove setIsAuthenticated false from sign out

demonzoo:
## Summary

Removed `setIsAuthenticated(false)` from `signOut` method in React and Vue SDKs, for it will cause an immediate state change in the client app, which might lead to problems that prevent browser navigating to the end session endpoint.

## Testing

- [x] Tested locally in AC, and the sign out worked as expected


simeng-li (approved):
LGTM

wangsijie (approved):

PR #308, opened 2022-06-14
chore: remove only-allow pnpm from all sub-packages

demonzoo:
## Summary

Remove `npx only-allow pnpm` from all sub-packages, leaving it only in the root package.json of our monorepo.

## Linear Issue Reference

[LOG-3040](https://linear.app/silverhand/issue/LOG-3040/support-npm-i-for-logto-js-sdks)

## Testing

N/A


xiaoyijun (approved):
LGTM

PR #311, opened 2022-06-27
feat(js,browser): configurable prompt

IceHe:
## Summary

`prompt` value can be configurable on our SDK.

To change the refresh token in the `CodeTokenResponse` to be optional for we won't get a refresh token from it if the prompt value is `login`. See Slack thread [discussion](https://silverhand-io.slack.com/archives/C02A8G4HVAM/p1649906832335309)

## Linear Issue Reference

LOG-2936

## Testing

image
image

### DIFF
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -1,4 +1,4 @@
-import { generateSignInUri, PromptValue } from '@logto/js';
+import { generateSignInUri, Prompt } from '@logto/js';
 import { Nullable } from '@silverhand/essentials';
 
 import LogtoClient, { AccessToken, LogtoClientError, LogtoConfig, LogtoSignInSessionItem } from '.';
@@ -127,12 +127,12 @@ describe('LogtoClient', () => {
 
     it('should use the default prompt value "consent" if we does not provide the custom prompt', () => {
       const logtoClient = new LogtoClientSignInSessionAccessor({ endpoint, appId }, requester);
-      expect(logtoClient.getLogtoConfig()).toHaveProperty('prompt', PromptValue.Consent);
+      expect(logtoClient.getLogtoConfig()).toHaveProperty('prompt', Prompt.Consent);
     });
 
     it('should use the custom prompt value "login"', () => {
       const logtoClient = new LogtoClientSignInSessionAccessor(
-        { endpoint, appId, prompt: PromptValue.Login },
+        { endpoint, appId, prompt: Prompt.Login },
         requester
       );
       expect(logtoClient.getLogtoConfig()).toHaveProperty('prompt', 'login');
--- a/packages/browser/src/index.test.ts
+++ b/packages/browser/src/index.test.ts
@@ -1,4 +1,4 @@
-import { generateSignInUri } from '@logto/js';
+import { generateSignInUri, PromptValue } from '@logto/js';
 import { Nullable } from '@silverhand/essentials';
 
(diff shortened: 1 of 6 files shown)


packages/js/src/consts/index.ts
```diff
@@ -28,3 +28,8 @@ export enum QueryKey {
   State = 'state',
   Token = 'token',
 }
+
+export enum PromptValue {
```
demonzoo, line 32: Just `Prompt`

packages/js/src/core/sign-in.ts
```diff
@@ -23,6 +23,7 @@ export const generateSignInUri = ({
   state,
   scopes,
   resources,
+  prompt,
```
demonzoo, line 26: `prompt = Prompt.Consent,`

packages/browser/src/index.ts
```diff
@@ -75,6 +77,7 @@ export default class LogtoClient {
   constructor(logtoConfig: LogtoConfig, requester = createRequester()) {
     this.logtoConfig = {
       ...logtoConfig,
+      prompt: logtoConfig.prompt ?? PromptValue.Consent,
```
demonzoo, line 80: ?? => `||`

packages/browser/src/index.ts
```diff
@@ -375,7 +378,7 @@ export default class LogtoClient {
     accessToken,
     expiresIn,
   }: CodeTokenResponse) {
-    this.refreshToken = refreshToken;
+    this.refreshToken = refreshToken ?? null;
```
demonzoo, line 381: Kinda forget. Do we have to use `null`?

packages/browser/src/index.test.ts
```diff
@@ -124,6 +124,19 @@ describe('LogtoClient', () => {
         'foo',
       ]);
     });
+
```
demonzoo, line 127: Would be nice if we can add a test case here, to verify that using prompt login can actually get an access token

packages/js/src/consts/index.ts
IceHe: Aligned with the name in Kotlin SDK.
I'm good with both.

packages/browser/src/index.ts
IceHe: refreshToken is `Nullable<…>`. Have to convert `undefined` to `null`

packages/js/src/consts/index.ts
IceHe: Updated to `Prompt`

packages/browser/src/index.test.ts
IceHe: Since UTs do not call actual Logto Core API, we cannot know whether the prompt value affects the response here.

packages/browser/src/index.ts
IceHe: since logtoConfig.prompt is the enum or undefined, keep using `??`

packages/js/src/core/sign-in.ts
IceHe: Updated. PTAL

```ts
export const generateSignInUri = ({
  authorizationEndpoint,
  clientId,
  redirectUri,
  codeChallenge,
  state,
  scopes,
  resources,
  prompt,
}: SignInUriParameters) => {
  const urlSearchParameters = new URLSearchParams({
    ……
    [QueryKey.Prompt]: prompt ?? Prompt.Consent,
    ……
  });


```

demonzoo (approved):

PR #316, opened 2022-06-29
fix(js): update userinfo response

gao-sun:
## Summary

update type per id token claims update

## Testing

CI

### DIFF
--- a/packages/js/src/core/user-info.ts
+++ b/packages/js/src/core/user-info.ts
@@ -1,11 +1,13 @@
+import { Nullable } from '@silverhand/essentials';
+
 import { Requester } from '../utils';
 
 export type UserInfoResponse = {
   sub: string;
-  name?: string;
-  username?: string;
-  avatar?: string;
-  role_names?: string[];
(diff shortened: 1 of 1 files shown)


packages/js/src/core/user-info.ts
```diff
@@ -1,11 +1,13 @@
...
 export type UserInfoResponse = {
   sub: string;
-  name?: string;
-  username?: string;
-  avatar?: string;
-  role_names?: string[];
+  name?: Nullable<string>;
+  username?: Nullable<string>;
+  avatar?: Nullable<string>;
+  role_names?: Nullable<string[]>;
```
IceHe, line 10: just wondering: they're reasonable to be `undefined` or `null`?

IceHe (approved):
Overall LGTM

packages/js/src/core/user-info.ts
gao-sun: userinfo will not have these fields when not corresponding scope provided

PR #319, opened 2022-06-30
fix(browser): should use prompt param on sign-in

demonzoo:
## Summary

Should use user specified prompt type in logtoClient when calling `.signIn()` method

## Testing

- [x] Unit test case added, all passed
- [x] Tested locally with React sample. Login was successful, and there was no `refreshToken` in localStorage


gao-sun (approved):

IceHe (approved):
LGTM

PR #321, opened 2022-07-06
fix: packing

gao-sun:
## Summary

- upgrade parcel
- export both esm and cjs

## Testing

create-react-app works fine with `@logto/js` imported


packages/browser/package.json
```diff
@@ -16,7 +21,7 @@
   "scripts": {
     "dev:tsc": "tsc -p tsconfig.build.json -w --preserveWatchOutput",
     "precommit": "lint-staged",
-    "build": "rm -rf lib/ && tsc -p tsconfig.build.json",
+    "build": "rm -rf lib/ && parcel build",
```
demonzoo, line 24: nit: Is `tsconfig.build.json` still needed? Seems not

demonzoo (approved):

packages/browser/package.json
gao-sun: we can use it for type check

gao-sun: removed

PR #329, opened 2022-07-11
feat(client): add client package

wangsijie:
## Summary

Extract 3 platform-related parts from `@logto/browser` and get a new package `@logto/client`:

1. requester
2. storage
3. handleRedirect

Next action: refactor `@logto/browser` to extend `@logto/client`.

## Testing

Passed all UTs.

### DIFF
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -1,0 +1,73 @@
+{
+  "name": "@logto/client",
+  "version": "1.0.0-alpha.2",
+  "source": "./src/index.ts",
+  "main": "./lib/index.js",
+  "exports": {
+    "require": "./lib/index.js",
+    "import": "./lib/module.js"
+  },
+  "module": "./lib/module.js",
+  "types": "./lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/logto-io/js.git",
+    "directory": "packages/client"
+  },
+  "scripts": {
+    "dev:tsc": "tsc -p tsconfig.build.json -w --preserveWatchOutput",
+    "precommit": "lint-staged",
+    "check": "tsc --noEmit",
+    "build": "rm -rf lib/ && pnpm check && parcel build",
+    "lint": "eslint --ext .ts src",
+    "test": "jest",
+    "test:coverage": "jest --silent --env=jsdom && jest --silent --coverage",
+    "prepack": "pnpm test"
+  },
+  "dependencies": {
+    "@logto/js": "^1.0.0-alpha.2",
+    "@silverhand/essentials": "^1.1.6",
+    "camelcase-keys": "^7.0.1",
+    "jose": "^4.3.8",
+    "js-base64": "^3.7.2",
+    "lodash.get": "^4.4.2",
+    "lodash.once": "^4.1.1",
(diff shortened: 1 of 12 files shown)


packages/client/package.json
```diff
@@ -0,0 +1,73 @@
...
+  },
+  "module": "./lib/module.js",
+  "types": "./lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/logto-io/js.git",
```
demonzoo, line 18:
```suggestion
    "url": "https://github.com/logto-io/client.git",
```

xiaoyijun (commented):
Wondering do we miss the `crypto` part? also `TextEncoder`.

packages/client/src/index.ts
```diff
@@ -0,0 +1,378 @@
...
+  private readonly getAccessTokenPromiseMap = new Map<string, Promise<string>>();
+  private _idToken: Nullable<string>;
+
+  constructor(
+    logtoConfig: LogtoConfig,
+    {
+      requester,
+      storage,
+      handleRedirect,
+    }: { requester: Requester; storage: Storage; handleRedirect: HandleRedirect }
```
demonzoo, line 78: nit: Maybe we should include logtoConfig in this new object as well? And it would be better to define a type name for it

demonzoo (approved):

wangsijie:
@xiaoyijun yes, crypto and textEncoder is in LOG-3616

packages/client/src/index.ts
wangsijie: I am gonna name it as something like "adapters", so I put logtoConfig outside the object. Good call, I'll think about it after finish crypto things.

packages/client/src/index.ts
```diff
@@ -0,0 +1,378 @@
...
+  }
+
+  private async _getOidcConfig() {
+    const { endpoint } = this.logtoConfig;
+    const discoveryEndpoint = getDiscoveryEndpoint(endpoint);
+
+    return fetchOidcConfig(discoveryEndpoint, this.requester);
+  }
+
+  private async _getJwtVerifyGetKey() {
```
simeng-li, line 345: nit: JwtVerifyGetKey better name?

packages/client/src/index.ts
```diff
@@ -0,0 +1,378 @@
...
+    }: { requester: Requester; storage: Storage; handleRedirect: HandleRedirect }
+  ) {
+    this.logtoConfig = {
+      ...logtoConfig,
+      prompt: logtoConfig.prompt ?? Prompt.Consent,
+      scopes: withReservedScopes(logtoConfig.scopes).split(' '),
+    };
+    this.storage = storage;
+    this.requester = requester;
+    this.handleRedirect = handleRedirect;
```
simeng-li, line 87: nit: just `navigate`?  handleRedirect feels more like some logic handler after redirect?

packages/client/src/index.ts
IceHe: follow the name from `jose`
image

packages/client/src/index.ts
wangsijie: Good call, but I may rename it later, to avoid merge confict.

PR #331, opened 2022-07-12
refactor(js,client,browser): move generators into adapters

wangsijie:
## Summary

Generators are based on `crypto` and `textEncoder`, and they are differ in node and browser.

Move them out from `js`, they are now "adapters" and the original code are moved to `browser`.

A note on lock file changes: some packages are upgrated during the moving from `js` to `browser`, cause large amount of hash changes.

## Testing

UTs.


packages/client/package.json
```diff
@@ -55,7 +54,6 @@
     "jest-matcher-specific-error": "^1.0.0",
     "lint-staged": "^13.0.0",
     "nock": "^13.1.3",
-    "node-fetch": "^2.6.7",
```
demonzoo, line 58: Should also remove `@peculiar/webcrypto`

packages/client/src/index.ts
```diff
@@ -54,47 +50,34 @@ export const LogtoSignInSessionItemSchema = type({
...
-
 export default class LogtoClient {
   protected readonly logtoConfig: LogtoConfig;
   protected readonly getOidcConfig = once(this._getOidcConfig);
   protected readonly getJwtVerifyGetKey = once(this._getJwtVerifyGetKey);
 
-  protected readonly storage: Storage;
-  protected readonly requester: Requester;
-  protected readonly handleRedirect: HandleRedirect;
+  protected readonly adapters: ClientAdapters;
```
demonzoo, line 58: nit: should use singular form `adapter: ClientAdapter`, because we didn't use plural form for `LogtoConfig`, either

packages/client/package.json
wangsijie: updated

packages/client/src/index.ts
wangsijie: updated

demonzoo (approved):
LGTM. Please don't forget to test with browser sample

gao-sun (approved):
closer to next.js :-P

PR #332, opened 2022-07-12
refactor(js,browser): reuse createRequester

wangsijie:
## Summary

Previously, I removed `createRequester` from `js`, but then realized it is useful, revert the deletion, and reuse this function in `browser`.

Due to parcel issue, `createRequester` is moved to `client` as a util function.

## Testing

UTs


packages/browser/jest.setup.js
```diff
@@ -4,6 +4,7 @@
 const crypto = require('crypto');
 
 const { location } = require('jest-location-mock');
+const fetch = require('node-fetch');
```
demonzoo, line 7: Do we still need `fetch` function in browser test? I thought the request related tests are located in `@logto/client`?

demonzoo (approved):

packages/browser/jest.setup.js
wangsijie: No, they are located in `browser`, in `client`, we use `jest.fn` to test.

PR #333, opened 2022-07-12
test(browser): storage test

wangsijie:
## Summary

Add unit tests for "BrowserStorage", then the old mockups can be removed, also `@logto/js` is removed from dev dependencies.

## Testing

UTs


packages/browser/package.json
```diff
@@ -30,11 +30,11 @@
   },
   "dependencies": {
     "@logto/client": "^1.0.0-alpha.2",
-    "@silverhand/essentials": "^1.1.6"
+    "@silverhand/essentials": "^1.1.6",
```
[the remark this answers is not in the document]
wangsijie: Oh, rebased with wrong branch, fixed

packages/browser/src/storage.ts
```diff
@@ -0,0 +1,38 @@
...
+  constructor(appId: string) {
+    this.storageKey = `${logtoStorageItemKeyPrefix}:${appId}`;
+  }
+
+  getItem(key: StorageKey): Nullable<string> {
+    if (key === 'signInSession') {
+      return sessionStorage.getItem(this.storageKey);
+    }
+
+    return localStorage.getItem(`${this.storageKey}:${key}`);
```
IceHe, line 18: may be better? and so do below
```suggestion
    const finalKey = key === 'signInSession'
      ? this.storageKey : `${this.storageKey}:${key}`

    return localStorage.getItem(finalKey);
```

packages/browser/src/storage.test.ts
```diff
@@ -0,0 +1,50 @@
+import { BrowserStorage, logtoStorageItemKeyPrefix } from './storage';
+
+describe('BrowserStorage', () => {
+  beforeEach(() => {
```
IceHe, line 4: clear in `afterEach` (discussed and aligned before (proposed by yijun))

packages/browser/src/index.test.ts
```diff
@@ -3,109 +3,8 @@ import LogtoClient from '.';
...
-    });
-  });
-
-  describe('handleRedirect', () => {
-    it('should cause location change after calling signIn', async () => {
-      expect(window.location.toString()).not.toContain(endpoint);
-      const logtoClient = new LogtoClient({ endpoint, appId });
-      await logtoClient.signIn(redirectUri);
-      expect(window.location.toString()).toContain(endpoint);
-    });
```
IceHe, line 109: this part is not related to storage, and should keep it?
(removed it since it broke UT for now?)

packages/browser/src/storage.ts
wangsijie: `signInSession` is stored in `sessionStorage`.

packages/browser/src/index.test.ts
wangsijie: You are right, this part is not related to storage, but still SHOULD be removed, since it requires some mocks in `@logto/js`, and another purpose of this PR is to remove `@logto/js`.

gao-sun (approved):

PR #338, opened 2022-07-13
feat(node): node sdk

wangsijie:
## Summary

New package: node

Extended from `client`, implement 4 adapters: `requester`, `generateCodeChallenge`, `generateCodeVerifier`, `generateState`.

Leave `storage` and `navigate` to users.

## Testing

UTs.

### DIFF
--- a/packages/node/src/include.d/node-fetch.d.ts
+++ b/packages/node/src/include.d/node-fetch.d.ts
@@ -1,0 +1,4 @@
+declare module 'node-fetch' {
+  const nodeFetch: typeof fetch;
+  export = nodeFetch;
+}
--- a/packages/node/tsconfig.json
+++ b/packages/node/tsconfig.json
@@ -1,0 +1,14 @@
+{
+    "extends": "@silverhand/ts-config/tsconfig.base",
+    "compilerOptions": {
+        "outDir": "lib",
+        "types": [
+            "jest",
+            "jest-matcher-specific-error"
+        ]
+    },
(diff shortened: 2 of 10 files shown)


packages/node/src/include.d/node-fetch.d.ts
```diff
@@ -0,0 +1,4 @@
+declare module 'node-fetch' {
+  const foo: typeof fetch;
+  export = foo;
```
gao-sun, line 3:
```suggestion
  const nodeFetch: typeof fetch;
  export = nodeFetch;
```

packages/node/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "extends": "@silverhand/ts-config/tsconfig.base",
+    "compilerOptions": {
+        "outDir": "lib",
+        "types": [
+            "jest",
+            "jest-matcher-specific-error"
```
demonzoo, line 7: nit: don't we need `node` here?

gao-sun (approved):

demonzoo (approved):
LGTM

packages/node/src/include.d/node-fetch.d.ts
demonzoo: +1

packages/node/tsconfig.json
wangsijie: is OK for now

PR #339, opened 2022-07-13
feat(next): init and add sign in route handler

wangsijie:
## Summary

Implement "Sign In" handler, when visiting "/api/sign-in", the user will be redirect to generated sign in uri (to Logto).

check "next-sample" for usage, if you are not familar with NextJS.

## Testing

UTs


packages/next/tsconfig.json
```diff
@@ -0,0 +1,14 @@
+{
+    "extends": "@silverhand/ts-config/tsconfig.base",
```
demonzoo, line 2: Extends from `ts-config-react` maybe?

packages/next/src/index.ts
```diff
@@ -0,0 +1,46 @@
...
+  private withIronSession(handler: NextApiHandler) {
+    return withIronSessionApiRoute(handler, {
+      cookieName: `logto:${this.config.appId}`,
+      password: this.config.cookieSecret,
+      cookieOptions: {
+        secure: this.config.cookieSecure,
+      },
+    });
+  }
+}
```
demonzoo, line 46: Can we provide hooks instead of class?

packages/next-sample/.eslintrc
```diff
@@ -0,0 +1,7 @@
+{
+  "extends": "@silverhand/react",
+  "rules": {
+    "consistent-default-export-name/default-export-match-filename": "off",
+    "react/react-in-jsx-scope": "off"
```
gao-sun, line 5: We can remove this line if we upgrade configs later

packages/next/package.json
```diff
@@ -0,0 +1,68 @@
...
+  "publishConfig": {
+    "access": "public"
+  },
+  "targets": {
+    "main": {
+      "context": "node",
+      "includeNodeModules": false
+    }
+  }
+}
```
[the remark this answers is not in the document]
gao-sun: Use `^16` instead since major version have breaking changes

packages/next/src/index.ts
gao-sun: It's server side so I guess nothing related to React?

packages/next/tsconfig.json
gao-sun: Ditto

packages/next-sample/libraries/logto.ts
```diff
@@ -0,0 +1,13 @@
+import NodeClient from '@logto/next';
```
gao-sun, line 1:
```suggestion
import LogtoClient from '@logto/next';
```

packages/next-sample/libraries/logto.ts
```diff
@@ -0,0 +1,13 @@
...
+
+export const logtoClient = new NodeClient({
+  appId: 'foo',
+  endpoint: 'https://logto.dev',
+  baseUrl: 'http://localhost:3000',
+  signInEndpoint: '/api/sign-in',
+  signInCallbackEndpoint: '/api/sign-in-callback',
+  signOutEndpoint: '/api/sign-out',
+  userEndpoint: '/api/user',
```
gao-sun, line 10: i think this is too much configs. maybe only `baseUrl` is required while other endpoints has default values? e.g.
```ts
{
  baseUrl: 'http://localhost:3000/api',
  // other endpoints are optional
}
```

gao-sun: or we can add these options until we need them?

packages/next/src/index.ts
demonzoo: Nah, this is Next.js SDK, and since Next is a `React framework` that build web app on the node server, it uses almost every concept in React, including JSX, React Dom, etc...
So I would assume the Next.js SDK would be much similar as React SDK

wangsijie: Is all been done in server side, but we can still provide a hook like "useLogto", and fetch info from an api endpoint later.

But this class is necessary.

packages/next-sample/libraries/logto.ts
xiaoyijun: Here may be confusing, for we have a `LogtoClient` in next SDK, too.

packages/next/package.json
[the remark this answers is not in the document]
wangsijie: updated

packages/next/tsconfig.json
wangsijie: updated

packages/next-sample/.eslintrc
wangsijie: unfortunately is still needed, can we look into this later?

packages/next-sample/libraries/logto.ts
wangsijie: removed.

packages/next-sample/.eslintrc
gao-sun: yes

packages/next-sample/tsconfig.json
```diff
@@ -0,0 +1,16 @@
+{
+  "extends": "@silverhand/ts-config-react/tsconfig.base",
+  "compilerOptions": {
+    "jsx": "preserve",
+    "allowJs": true,
```
gao-sun, line 5: no need for this?

packages/next/package.json
```diff
@@ -0,0 +1,72 @@
...
+    "extends": "@silverhand"
+  },
+  "prettier": "@silverhand/eslint-config/.prettierrc",
+  "publishConfig": {
+    "access": "public"
+  },
+  "targets": {
+    "main": {
+      "context": "node",
+      "includeNodeModules": false,
```
gao-sun, line 66: no need for this, see https://parceljs.org/features/targets/#includenodemodules

> The default is true for browser targets, and false for library targets.

packages/next/src/storage.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  getItem(key: StorageKey) {
+    const value = this.request.session[key];
+
+    if (typeof value === 'string') {
+      return value;
+    }
+
+    return null;
```
gao-sun, line 19: will it be better to return `String(value)` except `value === undefined`? e.g.

```ts
if (value === undefined) {
   return;
}

return String(value);
```

packages/next/src/storage.ts
```diff
@@ -0,0 +1,29 @@
...
+  }
+
+  removeItem(key: StorageKey) {
+    this.request.session[key] = undefined;
+  }
+
+  async save() {
```
gao-sun, line 26: will it be better to call `save()` inside the functions above?

wangsijie: That will need to change `setItem` to `async`, per offline discussion, consider refactor it later.

packages/next/src/storage.ts
wangsijie: updated

packages/next/package.json
wangsijie: updated

packages/next-sample/tsconfig.json
wangsijie: updated

gao-sun (approved):
good @demonzoo can double check if needed.

PR #355, opened 2022-07-18
feat(next): with logto api route

wangsijie:
## Summary

Add function: `withLogtoApiRoute`: read and assign auth state to `request`.

## Testing

UTs and sample.

### DIFF
--- a/packages/next-sample/libraries/fetch-json.ts
+++ b/packages/next-sample/libraries/fetch-json.ts
@@ -1,0 +1,14 @@
+export default async function fetchJson<JSON = unknown>(
+  input: RequestInfo,
+  init?: RequestInit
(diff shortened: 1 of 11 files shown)


packages/next-sample/libraries/fetch-json.ts
```diff
@@ -0,0 +1,14 @@
+export default async function fetchJson<JSON = unknown>(
+  input: RequestInfo,
+  init?: RequestInit
+): Promise<JSON> {
+  const response = await fetch(input, init);
+
+  const data = (await response.json()) as JSON;
```
demonzoo, line 7: Is it possible to avoid using `as`?

demonzoo (approved):

PR #358, opened 2022-07-19
feat(next): sign out

wangsijie:
## Summary

Add sign out route handler.

## Testing

UTs and sample(local tested)

### DIFF
--- a/packages/next/src/index.test.ts
+++ b/packages/next/src/index.test.ts
@@ -23,6 +23,7 @@ const handleSignInCallback = jest.fn();
 const getIdTokenClaims = jest.fn(() => ({
   sub: 'user_id',
 }));
+const signOut = jest.fn();
 
 jest.mock('./storage', () =>
   jest.fn(() => ({
@@ -48,6 +49,10 @@ jest.mock('@logto/node', () =>
     handleSignInCallback,
     isAuthenticated: true,
     getIdTokenClaims,
+    signOut: () => {
+      navigate(configs.baseUrl);
+      signOut();
+    },
   }))
 );
 
@@ -107,6 +112,23 @@ describe('Next', () => {
         },
       });
       expect(getIdTokenClaims).toHaveBeenCalled();
+    });
+  });
+
+  describe('handleSignOut', () => {
+    it('should redirect to Logto sign out url', async () => {
+      const client = new LogtoClient(configs);
+      await testApiHandler({
+        handler: client.handleSignOut(),
+        url: '/api/sign-out',
+        test: async ({ fetch }) => {
+          const response = await fetch({ method: 'GET', redirect: 'manual' });
+          const headers = response.headers as Map<string, string>;
+          expect(headers.get('location')).toEqual(`${configs.baseUrl}/`);
+        },
+      });
+      expect(save).toHaveBeenCalled();
+      expect(signOut).toHaveBeenCalled();
     });
   });
 });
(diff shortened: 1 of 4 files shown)


packages/next/src/index.ts
```diff
@@ -34,6 +34,19 @@ export default class LogtoClient {
       }
     });
 
+  handleSignOut = (redirectUri = this.config.baseUrl): NextApiHandler =>
+    this.withIronSession(async (request, response) => {
+      const nodeClient = this.createNodeClient(request);
+      await nodeClient.signOut(redirectUri);
+
+      request.session.destroy();
+      await this.storage?.save();
```
simeng-li, line 43: nit: just want to check we are not handling any exceptions here right? just throw whatever happens.

simeng-li (commented):
overall LGTM

simeng-li (approved):

packages/next/src/index.ts
wangsijie: YES, will be handled in LOG-3748, the final solution is not decided yet.

packages/next/src/index.ts
```diff
@@ -12,7 +12,7 @@ export default class LogtoClient {
   private storage?: NextStorage;
   constructor(private readonly config: LogtoNextConfig) {}
 
-  handleSignIn = (redirectUri = `${this.config.baseUrl}/api/sign-in`): NextApiHandler =>
+  handleSignIn = (redirectUri = `${this.config.baseUrl}/api/sign-in-callback`): NextApiHandler =>
```
demonzoo, line 15: How about rename it to `handleSignInCallback`, in order to align with other SDKs?

packages/next/src/index.ts
```diff
@@ -34,6 +34,19 @@ export default class LogtoClient {
       }
     });
 
+  handleSignOut = (redirectUri = this.config.baseUrl): NextApiHandler =>
+    this.withIronSession(async (request, response) => {
+      const nodeClient = this.createNodeClient(request);
```
demonzoo, line 39: Do we have to create a new `nodeClient` every time we handle sign-in or sign-out? Is it possible to maintain a singleton instance?

packages/next/src/index.test.ts
```diff
@@ -109,4 +114,21 @@ describe('Next', () => {
...
   });
+
+  describe('handleSignOut', () => {
+    it('should redirect to Logto sign out url', async () => {
+      const client = new LogtoClient(configs);
+      await testApiHandler({
+        handler: client.handleSignOut(),
+        url: '/api/sign-out',
+        test: async ({ fetch }) => {
+          const response = await fetch({ method: 'GET', redirect: 'manual' });
```
demonzoo, line 125: Why is `redirect: 'manual'` needed in here?

packages/next/src/index.ts
wangsijie: we have both `handleSignIn` and `handleSignInCallback`.

demonzoo: Oh, OK

packages/next/src/index.ts
wangsijie: No, they need to get `request` first, actually, each request is independent, they won't share anything (in memory) between API requests.

packages/next/src/index.test.ts
wangsijie: Because the real sign out url is not available in test environment. We just need to check the url that is going to redirect is corret.

demonzoo: ok

PR #359, opened 2022-07-19
feat(client): persist access token

wangsijie:
## Summary

Add an optional feature: persist access token to storage.

Before this, access token is only saved in instance property, when creating a new instance, it doesn't have any access token, but it can reuse refresh token from storage, and then grant a new access token. This is fine for SPA, since it only have an instance when the user navigates between pages. But in other circumstances, especially in traditional app when LogtoClient is run in server side, it will create a new instance each time when receiving a request, that means there will be so many instances sharing the same storage. Then, it'll be good to reuse access token.

## Testing

UTs.

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -16,10 +16,17 @@ import {
 import { Nullable } from '@silverhand/essentials';
 import { createRemoteJWKSet } from 'jose';
 import once from 'lodash.once';
-import { assert, Infer, string, type } from 'superstruct';
+import { assert } from 'superstruct';
 
 import { ClientAdapter } from './adapter';
 import { LogtoClientError } from './errors';
+import {
+  AccessToken,
+  LogtoAccessTokenMapSchema,
+  LogtoConfig,
+  LogtoSignInSessionItem,
+  LogtoSignInSessionItemSchema,
+} from './types';
 import { buildAccessTokenKey, getDiscoveryEndpoint } from './utils';
 
 export type { IdTokenClaims, LogtoErrorCode } from '@logto/js';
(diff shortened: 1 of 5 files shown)


packages/client/src/index.ts
```diff
@@ -360,5 +350,39 @@ export default class LogtoClient {
...
+    const raw = this.adapter.storage.getItem('accessToken');
+
+    if (!raw) {
+      return;
+    }
+
+    try {
+      const json: unknown = JSON.parse(raw);
+      assert(json, LogtoAccessTokenMapSchema);
+      this.accessTokenMap.clear();
```
demonzoo, line 381: Should we merge instead of override?

wangsijie: I think override is OK since the storage is for current `clientId` only.

packages/next/src/index.ts
```diff
@@ -35,12 +35,18 @@ export default class LogtoClient {
...
     this.storage = new NextStorage(request);
 
-    return new NodeClient(this.config, {
-      storage: this.storage,
-      navigate: (url) => {
-        this.navigateUrl = url;
+    return new NodeClient(
+      {
+        ...this.config,
+        persistAccessToken: true,
```
demonzoo, line 41: Should we leave this to user decision? If we do this, even if the user specificly set it to false in his app config, we will not respect that and still force use `true` in our SDK

demonzoo (approved):

packages/next/src/index.ts
wangsijie: Oops, fixed.
Changed to default true, and the user can specificly set it to false

PR #362, opened 2022-07-20
feat(next-sample): implement nextjs sample

wangsijie:
## Summary

Implement full sample of Next.js

## Testing

Local tested.

image

### DIFF
--- a/packages/next-sample/pages/protected.tsx
+++ b/packages/next-sample/pages/protected.tsx
@@ -1,0 +1,37 @@
+import Link from 'next/link';
+// eslint-disable-next-line import/no-named-as-default
+import Router from 'next/router';
+import { useEffect } from 'react';
+import useSWR from 'swr';
+
+const Protected = () => {
+  const { data, error } = useSWR<{ data: string }, Error>('/api/protected-resource');
+
+  useEffect(() => {
+    if (error?.message === 'Unauthorized') {
+      // eslint-disable-next-line @silverhand/fp/no-mutating-methods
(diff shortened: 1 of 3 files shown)


packages/next-sample/pages/protected.tsx
```diff
@@ -0,0 +1,37 @@
+import Link from 'next/link';
+// eslint-disable-next-line import/no-named-as-default
```
demonzoo, line 2: Is this needed?

packages/next-sample/pages/protected.tsx
```diff
@@ -0,0 +1,37 @@
...
+import Router from 'next/router';
+import { useEffect } from 'react';
+import useSWR from 'swr';
+
+const Protected = () => {
+  const { data, error } = useSWR<{ data: string }, Error>('/api/protected-resource');
+
+  useEffect(() => {
+    if (error?.message === 'Unauthorized') {
```
demonzoo, line 11: Use error status code instead?

demonzoo (approved):

packages/next-sample/pages/protected.tsx
wangsijie: Yes, "next/router" is odd, not go deep into it since is sample.

packages/next-sample/pages/protected.tsx
wangsijie: A ha, I wrote a very basic fetcher which doesn't throw "status code".

PR #363, opened 2022-07-20
feat(next): ssr support

wangsijie:
## Summary

Add "withLogtoSsr" to support [getServerSideProps](https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props)

## Testing

Local tested with sample.

### DIFF
--- a/packages/next-sample/pages/profile-ssr.tsx
+++ b/packages/next-sample/pages/profile-ssr.tsx
@@ -1,0 +1,64 @@
+import { LogtoUser } from '@logto/next';
+import { useMemo } from 'react';
+
+import { logtoClient } from '../libraries/logto';
+
+type Props = {
+  user: LogtoUser;
+};
+
+const ProfileSsr = ({ user }: Props) => {
+  const userInfo = useMemo(() => {
+    if (!user.isAuthenticated || !user.claims) {
+      return null;
+    }
+
+    return (
+      <div>
+        <h2>User info:</h2>
+        <table>
+          <thead>
+            <tr>
+              <th>Name</th>
+              <th>Value</th>
+            </tr>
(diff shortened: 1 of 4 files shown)


packages/next-sample/pages/profile-ssr.tsx
```diff
@@ -0,0 +1,64 @@
...
+    res.setHeader('location', '/api/login');
+    // eslint-disable-next-line @silverhand/fp/no-mutation
+    res.statusCode = 302;
+    res.end();
+  }
+
+  return {
+    props: { user },
+  };
+});
```
demonzoo, line 64: 1. Maybe we should put this on top of the file?
2. Seems we don't need to `export`

demonzoo (approved):

packages/next-sample/pages/profile-ssr.tsx
wangsijie: export this function as a requirement of Next.js, https://nextjs.org/docs/basic-features/data-fetching/get-server-side-props

wangsijie: Also, put this after "default export(component)" is common.

packages/next/src/index.ts
```diff
@@ -97,14 +105,27 @@ export default class LogtoClient {
...
+    return {
       cookieName: `logto:${this.config.appId}`,
       password: this.config.cookieSecret,
       cookieOptions: {
         secure: this.config.cookieSecure,
         maxAge: 14 * 24 * 60 * 60,
       },
-    });
+    };
+  }
```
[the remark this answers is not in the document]
wangsijie: Of couse, updated.

PR #364, opened 2022-07-21
feat(next): set getAccessToken as an option

wangsijie:
## Summary

For checking isAuthenticated, there are 2 steps:

1. check `logtoClient.isAuthenticated`, that actually means that `idToken` is exist (session is exist).
2. try to get an accessToken, make sure the refresh token is still valid.

step 2 is optional and skiped by default.

## Testing

UTs and local tested with sample.

### DIFF
--- a/packages/next/src/types.ts
+++ b/packages/next/src/types.ts
@@ -28,4 +28,14 @@ export type LogtoNextConfig = LogtoConfig & {
 export type LogtoUser = {
   isAuthenticated: boolean;
   claims?: IdTokenClaims;
+  accessToken?: string;
+};
+
+/**
+ * @getAccessToken: if set to true, will try to get an access token and attach to req.user,
+ * if unable to grant an access token, will set req.user.isAuthenticated to false,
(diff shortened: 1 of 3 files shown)


packages/next/src/types.ts
```diff
@@ -28,4 +28,14 @@ export type LogtoNextConfig = LogtoConfig & {
 export type LogtoUser = {
   isAuthenticated: boolean;
   claims?: IdTokenClaims;
+  accessToken?: string;
+};
+
+/**
+ * @getAccessToken: if set to true, will try to get an access token and attach to req.user,
+ * if unable to grant and access token, will set req.user.isAuthenticated to false,
```
darcyYe, line 36:
```suggestion
 * if unable to grant an access token, will set req.user.isAuthenticated to false,
```

darcyYe (approved):

packages/next/src/types.ts
wangsijie: updated

PR #371, opened 2022-07-27
refactor(js,client): createRequester

IceHe:
## Summary

- Add `logtoRequestErrorSchema` to check Logto request error
- Refactor `createRequester`
    - Throw Logto request error when the response contains both `code` and `message`
    - Otherwise, throw Logto error with the message `Unexpected response error from the server.`

## Testing

image

image

- [x] Locally tested on Admin Console.

    Original

    image

    Now

    image

### DIFF
--- a/packages/client/src/utils/requester.test.ts
+++ b/packages/client/src/utils/requester.test.ts
@@ -1,4 +1,4 @@
-import { LogtoRequestError } from '@logto/js';
+import { LogtoError, LogtoRequestError } from '@logto/js';
 
 import { createRequester } from './requester';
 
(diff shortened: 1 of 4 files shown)


packages/client/src/utils/requester.test.ts
```diff
@@ -27,37 +28,49 @@ describe('createRequester', () => {
       await expect(requester('foo')).rejects.toMatchError(new LogtoRequestError(code, message));
     });
 
-    test('failing response json with only code should throw StructError', async () => {
+    test('failing response json with more than code and message should throw LogtoRequestError with same code and message', async () => {
```
wangsijie, line 31: nit: with additional info?

wangsijie (approved):
LGTM

PR #372, opened 2022-08-01
feat(node): support client secret

wangsijie:
## Summary

Add support for OIDC's `client_secret`, it is called `appSecret` here.

## Testing

Local tested.

### DIFF
--- a/packages/node/src/index.ts
+++ b/packages/node/src/index.ts
@@ -15,9 +15,27 @@ export { LogtoError, OidcError, Prompt, LogtoRequestError, LogtoClientError } fr
 
 export default class LogtoClient extends BaseClient {
   constructor(config: LogtoConfig, adapter: Pick<ClientAdapter, 'navigate' | 'storage'>) {
+    const { appId, appSecret } = config;
+
     super(config, {
       ...adapter,
-      requester: createRequester(fetch),
+      requester: createRequester(
+        appSecret
+          ? async (...args: Parameters<typeof fetch>) => {
+              const [input, init] = args;
+
+              return fetch(input, {
+                ...init,
+                headers: {
+                  ...init?.headers,
+                  authorization: `basic ${Buffer.from(`${appId}:${appSecret}`, 'utf8').toString(
+                    'base64'
+                  )}`,
+                },
+              });
+            }
+          : fetch
+      ),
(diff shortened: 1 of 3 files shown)


packages/node/src/index.ts
```diff
@@ -17,7 +17,22 @@ export default class LogtoClient extends BaseClient {
...
-      requester: createRequester(fetch),
+      requester: createRequester(
+        config.appSecret
+          ? async (...args: Parameters<typeof fetch>) => {
+              return fetch(args[0], {
+                ...args[1],
+                headers: {
+                  ...args[1]?.headers,
+                  authorization: `basic ${Buffer.from(
+                    `${config.appId}:${config.appSecret ?? ''}`,
```
darcyYe, line 28: Will config.appSecret be false as the entry condition is a truthy config.appSecret?

packages/node/src/index.ts
```diff
@@ -17,7 +17,22 @@ export default class LogtoClient extends BaseClient {
   constructor(config: LogtoConfig, adapter: Pick<ClientAdapter, 'navigate' | 'storage'>) {
     super(config, {
       ...adapter,
-      requester: createRequester(fetch),
+      requester: createRequester(
+        config.appSecret
+          ? async (...args: Parameters<typeof fetch>) => {
+              return fetch(args[0], {
+                ...args[1],
```
demonzoo, line 24: What are `args[0]` and `args[1]` specifically? `url` and `configs`?
Would it be better to de-construct them first?
```ts
const [url, configs] = args;
```

packages/node/src/index.ts
demonzoo: Would it be better to apply client secret authorization only when `appSecret` is not empty?
```
...conditional(config.appSecret && ({ authorization: `basic ${ ... }` }))
```

packages/node/src/index.ts
wangsijie: updated

image

packages/node/src/index.ts
wangsijie: in line 21:
```
config.appSecret
?
```

demonzoo (approved):

darcyYe (approved):

PR #375, opened 2022-08-03
feat(express): init express and add routes handler

wangsijie:
## Summary

Init express SDK and implement handle routes:

1. sign-in
2. sign-in-callback
3. sign-out

## Testing

UTs and local tested with real express app.

### DIFF
--- a/packages/express/package.json
+++ b/packages/express/package.json
@@ -1,0 +1,73 @@
+{
+  "name": "@logto/express",
+  "version": "1.0.0-beta.0",
+  "source": "./src/index.ts",
+  "main": "./lib/index.js",
+  "exports": {
+    "require": "./lib/index.js",
+    "import": "./lib/module.js"
+  },
+  "module": "./lib/module.js",
+  "types": "./lib/index.d.ts",
+  "files": [
+    "lib"
+  ],
+  "license": "MIT",
+  "repository": {
+    "type": "git",
+    "url": "https://github.com/logto-io/js.git",
+    "directory": "packages/express"
+  },
+  "scripts": {
+    "dev:tsc": "tsc -p tsconfig.build.json -w --preserveWatchOutput",
+    "precommit": "lint-staged",
+    "check": "tsc --noEmit",
+    "build": "rm -rf lib/ && pnpm check && parcel build",
+    "lint": "eslint --ext .ts src",
+    "test": "jest",
+    "test:coverage": "jest --silent --coverage",
+    "prepack": "pnpm test"
+  },
+  "dependencies": {
+    "@logto/node": "^1.0.0-beta.0"
+  },
+  "devDependencies": {
+    "@jest/types": "^27.5.1",
+    "@parcel/core": "^2.6.2",
+    "@parcel/packager-ts": "^2.6.2",
+    "@parcel/transformer-typescript-types": "^2.6.2",
+    "@silverhand/eslint-config": "^0.17.0",
+    "@silverhand/ts-config": "^0.14.0",
+    "@silverhand/ts-config-react": "^0.14.0",
+    "@types/express": "^4.17.13",
+    "@types/jest": "^27.4.0",
+    "eslint": "^8.9.0",
+    "express": "^4.18.1",
+    "jest": "^27.5.1",
+    "jest-location-mock": "^1.0.9",
+    "jest-matcher-specific-error": "^1.0.0",
+    "lint-staged": "^13.0.0",
+    "parcel": "^2.6.2",
+    "prettier": "^2.3.2",
+    "ts-jest": "^27.0.4",
+    "typescript": "^4.5.5"
+  },
+  "peerDependencies": {
+    "express": ">=4"
+  },
+  "eslintConfig": {
+    "extends": "@silverhand"
+  },
+  "prettier": "@silverhand/eslint-config/.prettierrc",
+  "publishConfig": {
+    "access": "public"
+  },
+  "targets": {
+    "main": {
+      "context": "node",
+      "engines": {
+        "node": "^16"
+      }
+    }
+  }
+}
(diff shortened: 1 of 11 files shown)


packages/express/src/index.ts
```diff
@@ -0,0 +1,93 @@
+import { IncomingMessage } from 'http';
+
+import NodeClient from '@logto/node';
+import { Request, Response, NextFunction, Router } from 'express';
+
+import ExpressStorage from './storage';
+import { LogtoExpressConfig } from './types';
```
demonzoo, line 7: Can we use a more generic type, like `LogtoConfig` or something?

packages/express/src/types.ts
```diff
@@ -0,0 +1,11 @@
+import { LogtoConfig } from '@logto/node';
+
+declare module 'http' {
+  interface IncomingMessage {
+    session: Record<string, string | undefined>;
+  }
+}
+
+export type LogtoExpressConfig = LogtoConfig & {
+  baseUrl: string;
```
demonzoo, line 10: Can we add `baseUrl` to `LogtoConfig` in node sdk? It seems it is also used in next.js sdk.

packages/express/src/index.test.ts
```diff
@@ -0,0 +1,103 @@
+import LogtoClient from '.';
+import { testMiddleware } from './test-utils';
+import { LogtoExpressConfig } from './types';
+
+const signInUrl = 'http://mock-logto-server.com/sign-in';
+
+const configs: LogtoExpressConfig = {
+  appId: 'app_id_value',
+  endpoint: 'https://logto.dev',
+  baseUrl: 'http://localhost:3000',
```
demonzoo, line 10: Don't we need to support `cookieSecret` in express sdk?

packages/express/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+        persistAccessToken: this.config.persistAccessToken ?? true,
+      },
+      {
+        storage,
+        navigate: (url) => {
+          response.redirect(url);
+        },
+      }
+    );
+  }
```
demonzoo, line 84: These logics are seen in Next.js SDK as well. So I'm curious if it is possible to put some of these logics in the parent Node.js SDK, in the form of an abstract class or interface?

packages/express/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+
+  private checkSession(request: IncomingMessage) {
+    // We assume that `session` is configured in the express app, but need to check it there.
+    // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+    if (!request.session) {
+      throw new Error('You should configure express-session before using Logto express SDK.');
```
demonzoo, line 90: Better throwing this error as a LogtoClientError with an error code, as one day we are going to i18n these error messages.

packages/express/src/index.ts
```diff
@@ -0,0 +1,93 @@
...
+      }
+
+      response.status(404).end();
+    });
+
+    return router;
+  };
+
+  private createNodeClient(request: IncomingMessage, response: Response) {
+    this.checkSession(request);
```
demonzoo, line 69: Just curious why don't we check session in Next.js SDK

packages/express/src/storage.ts
```diff
@@ -0,0 +1,25 @@
...
+    if (value === undefined) {
+      return null;
+    }
+
+    return String(value);
+  }
+
+  removeItem(key: StorageKey) {
+    this.request.session[key] = undefined;
+  }
```
demonzoo, line 24: Maybe a dumb question: why don't we have `save` function in express storage? Or maybe in the opposite way: why do we need `save` function in Next.js SDK?

packages/express/src/index.test.ts
wangsijie: In express, middlewares usually only do one thing, so we leave it to other middleware named "express-session", will mention this in docs.

packages/express/src/index.ts
wangsijie: If we look into details, we can find that thay are not the same, just looks like the same. But I will refactor and put redunctant logics to Node.js SDK if it appears, good point, thanks.

packages/express/src/index.ts
wangsijie: Just like the previous conversation said, our Next.js SDK contains cookie, but express does not.

packages/express/src/storage.ts
wangsijie: It is all because of the difference between 2 session plugins.

packages/express/src/index.ts
charIeszhao: Cool, thx

charIeszhao (approved):

packages/express/src/types.ts
wangsijie: But it doesn't make sense for Node SDK, because `baseUrl` is useless in node sdk.

charIeszhao: OK

packages/express/src/index.ts
wangsijie: nice, updated

packages/express/package.json
```diff
@@ -0,0 +1,73 @@
...
+  "dependencies": {
+    "@logto/node": "^1.0.0-beta.0"
+  },
+  "devDependencies": {
+    "@jest/types": "^27.5.1",
+    "@parcel/core": "^2.6.2",
+    "@parcel/packager-ts": "^2.6.2",
+    "@parcel/transformer-typescript-types": "^2.6.2",
+    "@silverhand/eslint-config": "^0.17.0",
+    "@silverhand/ts-config": "^0.14.0",
```
charIeszhao, line 40: use 1.0.0-rc.2

PR #380, opened 2022-08-10
feat(express-sample): add express sample code

wangsijie:
## Summary

Implement express sample.

## Testing

Local tested

### DIFF
--- a/packages/express-sample/package.json
+++ b/packages/express-sample/package.json
@@ -1,0 +1,38 @@
+{
+  "name": "@logto/express-sample",
+  "version": "1.0.0-beta.0",
+  "license": "MIT",
+  "private": true,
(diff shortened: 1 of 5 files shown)


packages/express-sample/package.json
```diff
@@ -0,0 +1,38 @@
+{
+  "name": "@logto/express-sample",
+  "version": "1.0.0-beta.0",
```
IceHe, line 3: better be `1.0.0-beta.3`?

charIeszhao: Nah, the latest version is `1.0.0-beta.2`, but it's also OK to use a smaller version number as it will be automatically bumped when publishing beta.3

charIeszhao (approved):

PR #382, opened 2022-08-11
refactor(express): extract class into functions

wangsijie:
## Summary

Remove "class", export 2 middlewares: handleAuthRoutes and withLogto

```
app.use(handleAuthRoutes(config));
app.use(withLogto(config));
```

## Testing

UTs and express sample.


packages/express-sample/src/app.ts
```diff
@@ -1,10 +1,16 @@
 import http from 'http';
 
+import { handleAuthRoutes, withLogto } from '@logto/express';
 import cookieParser from 'cookie-parser';
 import express, { Request, Response, NextFunction } from 'express';
 import session from 'express-session';
 
-import { logtoClient } from './logto';
+const config = {
```
charIeszhao, line 8: nit: `const config: LogtoConfig = { ... }`

packages/express/src/index.test.ts
```diff
@@ -56,56 +56,36 @@ describe('Express', () => {
...
-    it('should call client.handleSignInCallback and redirect to home', async () => {
-      const client = new LogtoClient(configs);
-      await testMiddleware({
-        middleware: client.handleSignInCallback(),
-        url: '/api/logto/sign-in-callback',
-        test: async ({ response }) => {
-          expect(response.redirect).toHaveBeenCalledWith(configs.baseUrl);
-        },
+    describe('handleSignInCallback', () => {
+      it('should call client.handleSignInCallback and redirect to home', async () => {
```
charIeszhao, line 69: nit: redirect home or redirect to home page

packages/express/src/index.ts
```diff
@@ -15,92 +15,78 @@ export type Middleware = (
...
-      const nodeClient = this.createNodeClient(request, response);
+const createNodeClient = (
+  request: IncomingMessage,
+  response: Response,
+  config: LogtoExpressConfig
+) => {
+  // We assume that `session` is configured in the express app, but need to check it there.
+  // eslint-disable-next-line @typescript-eslint/no-unnecessary-condition
+  if (!request.session) {
+    throw new LogtoExpressError('session_not_configured');
```
charIeszhao, line 26: nit: maybe `session not found`?

charIeszhao: Also why do we need to maintain a new error class `LogtoExpressError`? Can we re-use `LogtoClientError`?

charIeszhao (approved):

packages/express/src/index.ts
wangsijie: 1. `session not found` sounds like the user has no session, `not configured` is more accurate.
2. `LogtoClientError` don't have this error key.

charIeszhao: Well, OK

PR #388, opened 2022-08-19
refactor: make storage functions async

charIeszhao:
## Summary

Since React Native storage solutions are all async and we need to write React Native SDK to extend from @logto/client , it is required to implement the Storage interface in React Native.

However, the existing functions (getItem, setItem, removeItem) in Storage interface are all synchronous, we can't implement the interface with an asynchronous storage solution.

Therefore, we need to make storage functions async in @logto/client in order to unblock React Native SDK development.

## Testing

- [x] Build successfully
- [x] Sign-in and sign-out worked in all sample projects
- [x] Passed all test cases

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -40,13 +40,10 @@ export default class LogtoClient {
   protected readonly logtoConfig: LogtoConfig;
   protected readonly getOidcConfig = once(this._getOidcConfig);
   protected readonly getJwtVerifyGetKey = once(this._getJwtVerifyGetKey);
-
   protected readonly adapter: ClientAdapter;
-
   protected readonly accessTokenMap = new Map<string, AccessToken>();
 
   private readonly getAccessTokenPromiseMap = new Map<string, Promise<string>>();
-  private _idToken: Nullable<string>;
 
   constructor(logtoConfig: LogtoConfig, adapter: ClientAdapter) {
     this.logtoConfig = {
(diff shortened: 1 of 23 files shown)


packages/client/src/index.ts
```diff
@@ -55,78 +52,27 @@ export default class LogtoClient {
...
-  protected set signInSession(logtoSignInSessionItem: Nullable<LogtoSignInSessionItem>) {
-    if (!logtoSignInSessionItem) {
-      this.adapter.storage.removeItem('signInSession');
-
-      return;
-    }
-
-    const jsonItem = JSON.stringify(logtoSignInSessionItem);
-    this.adapter.storage.setItem('signInSession', jsonItem);
+  async getIsAuthenticated() {
```
wangsijie, line 61: maybe `async isAuthenticated()`?

packages/node/src/index.ts
```diff
@@ -18,22 +18,23 @@ export { LogtoError, OidcError, Prompt, LogtoRequestError, LogtoClientError } fr
 
 export default class LogtoClient extends BaseClient {
   constructor(config: LogtoConfig, adapter: Pick<ClientAdapter, 'navigate' | 'storage'>) {
-    const { appId, appSecret } = config;
-
     super(config, {
       ...adapter,
       requester: createRequester(
-        appSecret
+        config.appSecret
```
wangsijie, line 24: wrong base?

charIeszhao: There's a new lint rule that makes sure the `super` comes first in the constructor. Therefore we can't declare variables before the `super` call. Have to make this change to adopt the new rule.

packages/client/src/index.ts
charIeszhao: np, updated

wangsijie (approved):
LGTM

PR #390, opened 2022-08-24
feat(js,client): enable user endpoint in js core and client sdks

charIeszhao:
## Summary

Enable user endpoint in JS core and client SDKs

## Testing

- [x] Locally tested
- [x] Passed all UTs

### DIFF
--- a/packages/js/src/core/user-info.ts
+++ b/packages/js/src/core/user-info.ts
@@ -1,0 +1,29 @@
+import { Requester } from '../types';
+
+type Identity = {
+  userId: string;
+  details?: Record<string, unknown>;
+};
+
+export type UserInfoResponse = {
+  sub: string;
+  name?: string;
+  username?: string;
+  picture?: string;
+  role_names?: string[];
+  email?: string;
+  email_verified?: boolean;
+  phone_number?: string;
+  phone_number_verified?: boolean;
+  custom_data?: unknown;
+  identities?: Record<string, Identity>;
+};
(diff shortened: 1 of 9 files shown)


packages/js/src/core/user-info.ts
```diff
@@ -0,0 +1,18 @@
...
+  role_names?: string[];
+};
+
+export const fetchUserInfo = async (
+  userInfoEndpoint: string,
+  accessToken: string,
+  requester: Requester
+): Promise<UserInfoResponse> =>
+  requester<UserInfoResponse>(userInfoEndpoint, {
+    headers: { Authorization: `Bearer ${accessToken}` },
```
wangsijie, line 17: In node sdk, we mutate the "headers", would you mind checking if there is a conflict issue?

wangsijie: for appSecret

charIeszhao: Hmmm, I think you're right. Should move this to browser SDK

wangsijie: But seems OK for now, this endpoint does not require `appSecret` to work?

charIeszhao: No, it does not require appSecret. Yeah, I thought about this couple days ago but forgot to update the comment. So I guess it's fine for now

wangsijie (approved):
LGTM

PR #399, opened 2022-09-08
fix(react,vue): fix mis-handled isAuthenticated state in react and vue sdks

charIeszhao:
## Summary

Fixed the mis-handled `isAuthenticated` state in React and Vue SDKs, which was introduced in 1.0.0-beta.3 and has caused sign-in issues in admin console and demo-app.

## Testing

- [x] Locally tested with demo app and admin console, their sign-in flow works as expected

### DIFF
--- a/packages/react-sample/src/RequireAuth.ts
+++ b/packages/react-sample/src/RequireAuth.ts
@@ -4,13 +4,13 @@ import { useEffect } from 'react';
 import { redirectUrl } from './consts';
 
 const RequireAuth = ({ children }: { children: JSX.Element }) => {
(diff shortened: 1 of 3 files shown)


packages/react/src/context.tsx
```diff
@@ -18,7 +18,7 @@ export const throwContextError = (): never => {
 export const LogtoContext = createContext<LogtoContextProps>({
   logtoClient: undefined,
   isAuthenticated: false,
-  loadingCount: 0,
+  loadingCount: 1,
```
gao-sun, line 21:
```suggestion
  loadingCount: 0,
```

gao-sun (approved):

PR #406, opened 2022-09-19
fix: remove persist access token

wangsijie:
## Summary

Remove `persistAccessToken` from configs, it is no longer configuable.

## Testing

UTs


simeng-li (approved):
LGTM

PR #407, opened 2022-09-19
chore: update default scopes and IdTokenClaim definition

charIeszhao:
## Summary

* JS SDKs export `ReservedScope` and `UserScope` from @logto/core-kit
* By default, the SDK provides `openid`, `offline_access` and `profile` scopes
* Update the type definition of `IdTokenClaim`, change "avatar" to "picture", and add email and phone_number properties

## Testing

- [x] Tested in browser sample, react sample and vue sample projects, `fetchUserInfo` API can return corresponding user data structure by given different scopes. Working fine.
- [x] Tested in the above sample projects, that ID token claims are also affected by given scopes.

### DIFF
--- a/packages/browser-sample/src/index.js
+++ b/packages/browser-sample/src/index.js
@@ -1,4 +1,4 @@
-import LogtoClient from '@logto/browser';
+import LogtoClient, { UserScope } from '@logto/browser';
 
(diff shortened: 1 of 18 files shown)


simeng-li (approved):
LGTM

PR #408, opened 2022-09-20
refactor(js,client): remove superstruct

wangsijie:
## Summary

Remove superstruct, replaced with own type check functions.

The package "superstruct" is "es module" and can not be "required" in some circumstances, we can fix it by setting "alias", but our users will have that problem too.

## Testing

UTs.

### DIFF
--- a/packages/client/src/types/index.ts
+++ b/packages/client/src/types/index.ts
@@ -1,5 +1,4 @@
-import { Prompt } from '@logto/js';
-import { Infer, number, record, string, type } from 'superstruct';
+import { isArbitraryObject, Prompt } from '@logto/js';
 
 export type LogtoConfig = {
   endpoint: string;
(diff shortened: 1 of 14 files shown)


packages/client/src/types/index.ts
```diff
@@ -11,20 +10,46 @@ export type LogtoConfig = {
...
+      typeof value.token !== 'string' ||
+      typeof value.scope !== 'string' ||
+      typeof value.expiresAt !== 'number'
+    ) {
+      return false;
+    }
+  }
 
-export type LogtoSignInSessionItem = Infer<typeof LogtoSignInSessionItemSchema>;
+  return true;
```
charIeszhao, line 47: Direct return the result?
```
return typeof value.token !== 'string' || typeof value.scope !== 'string' || typeof value.expiresAt !== 'number';
```

packages/client/src/types/index.ts
```diff
@@ -11,20 +10,46 @@ export type LogtoConfig = {
...
 
-export const LogtoSignInSessionItemSchema = type({
-  redirectUri: string(),
-  codeVerifier: string(),
-  state: string(),
-});
+  for (const value of Object.values(data)) {
+    if (!isArbitraryObject(value)) {
+      return false;
+    }
```
charIeszhao, line 36: What if the map is empty?

packages/client/src/types/index.ts
wangsijie: updated with array.every

packages/client/src/types/index.ts
wangsijie: Empty map is valid?

gao-sun (approved):

charIeszhao (approved):

packages/client/src/types/index.ts
charIeszhao: Hmmm, you're right. No problem in here

PR #413, opened 2022-09-29
feat(express,next,node): support fetchUserInfo

wangsijie:
## Summary

Add config "fetchUserInfo" to `express` and `next`, enable user info endpoint in Node.js SDK family.

## Testing

Local tested with express-sample and next-sample.

image

### DIFF
--- a/packages/express-sample/src/app.ts
+++ b/packages/express-sample/src/app.ts
@@ -24,7 +24,6 @@ const app = express();
 app.use(cookieParser());
 app.use(session({ secret: 'keyboard cat', cookie: { maxAge: 14 * 24 * 60 * 60 } }));
 app.use(handleAuthRoutes(config));
-app.use(withLogto(config));
 
 app.get('/', (request, response) => {
   response.setHeader('content-type', 'text/html');
@@ -37,11 +36,23 @@ app.get('/', (request, response) => {
   );
 });
 
-app.get('/user', (request, response) => {
+app.get('/local-user-claims', withLogto(config), (request, response) => {
   response.json(request.user);
 });
 
-app.get('/protected', requireAuth, (request, response) => {
+app.get(
+  '/remote-full-user',
+  withLogto({
+    ...config,
+    // Fetch user info from remote, this may slowdown the response time, not recommended.
+    fetchUserInfo: true,
+  }),
+  (request, response) => {
+    response.json(request.user);
+  }
+);
+
+app.get('/protected', withLogto(config), requireAuth, (request, response) => {
   response.end('protected resource');
 });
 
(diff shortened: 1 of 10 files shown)


packages/node/src/index.ts
```diff
@@ -70,6 +70,7 @@ export default class LogtoClient extends BaseClient {
       return {
         isAuthenticated,
         claims,
+        userInfo: await this._fetchUserInfo(fetchUserInfo),
```
charIeszhao, line 73: How about checking and calling `fetchUserInfo` in prior? e.g.
```ts
const userInfo = conditional(fetchUserInfo && (await this.fetchUserInfo()));
```
This way, you don't have to declare a private `_fetchUserInfo` method

packages/express-sample/src/app.ts
```diff
@@ -37,11 +36,23 @@ app.get('/', (request, response) => {
...
 });
 
-app.get('/user', (request, response) => {
+app.get('/user', withLogto(config), (request, response) => {
   response.json(request.user);
 });
 
-app.get('/protected', requireAuth, (request, response) => {
+app.get(
+  '/user-fetched',
```
charIeszhao, line 44: Nit: Maybe `full-user` or `remote-user`?

wangsijie: updated

packages/node/src/index.ts
wangsijie: updated

packages/next-sample/pages/index.tsx
```diff
@@ -5,16 +5,17 @@ import useSWR from 'swr';
 
 const Home = () => {
   const { data } = useSWR<LogtoContext>('/api/logto/user');
+  const { data: dataWithUserInfo } = useSWR<LogtoContext>('/api/logto/user-info');
```
charIeszhao, line 8: Seems we also need to update these two routes

packages/node/src/index.ts
```diff
@@ -79,12 +82,14 @@ export default class LogtoClient extends BaseClient {
...
-    } catch {
-      return {
-        isAuthenticated: false,
-      };
-    }
+    } catch {}
+
+    return {
+      isAuthenticated: false,
+    };
```
charIeszhao, line 92: Use `finally` instead?

charIeszhao (approved):

packages/next-sample/pages/index.tsx
wangsijie: `/api/logto/user` is the SDK's built-in route (also recommended), so I don't change the name, instead, I add 2 comments on them.

PR #417, opened 2022-10-14
feat(remix): remix SDK

akoenig:
## Summary

This PR integrates the Remix SDK for Logto originally developed in [openformation/logto-remix](https://openformation/logto-remix).

## Testing

Created a demo application at [openformation/logto-remix](https://openformation/logto-remix).

### DIFF
--- a/packages/remix/src/infrastructure/logto/create-client.ts
+++ b/packages/remix/src/infrastructure/logto/create-client.ts
@@ -1,0 +1,16 @@
+import LogtoClient, { LogtoConfig } from '@logto/node';
+
+import { LogtoStorage } from './create-storage';
+
+export const makeLogtoClient =
+  (config: LogtoConfig, storage: LogtoStorage) =>
+  // Have to deactivate the eslint rule here as the `LogtoClient`
+  // awaits a `navigate` function.
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  (navigate: (url: string) => void = () => {}) => {
+    return new LogtoClient(config, { storage, navigate });
+  };
+
+export type CreateLogtoClient = ReturnType<typeof makeLogtoClient>;
+
+export { type LogtoConfig } from '@logto/node';
--- a/packages/remix/src/infrastructure/logto/create-storage.spec.ts
+++ b/packages/remix/src/infrastructure/logto/create-storage.spec.ts
@@ -1,0 +1,53 @@
+import { createSession } from '@remix-run/node';
+
+import { createStorage } from './create-storage';
+
+describe('infrastructure:logto:createStorage', () => {
+  it('can create a LogtoStorage instance', async () => {
+    const session = createSession();
+    const storage = createStorage(session);
+
+    expect(storage.constructor.name).toBe('LogtoStorage');
+  });
+
+  it('can set items', async () => {
+    const session = createSession();
+    const storage = createStorage(session);
+
+    await storage.setItem('idToken', 'a');
+    await storage.setItem('accessToken', 'b');
+    await storage.setItem('refreshToken', 'c');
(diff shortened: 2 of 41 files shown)


packages/remix/src/infrastructure/logto/create-client.ts
```diff
@@ -0,0 +1,16 @@
...
+import { LogtoStorage } from './create-storage';
+
+export const makeLogtoClient =
+  (config: LogtoConfig, storage: LogtoStorage) =>
+  // Have to deactivate the eslint rule here as the `LogtoClient`
+  // awaits a `navigate` function.
+  // eslint-disable-next-line @typescript-eslint/no-empty-function
+  (navigate: (url: string) => void = () => {}) => {
+    return new LogtoClient(config, { storage, navigate });
+  };
```
charIeszhao, line 12: I'm just wondering how Remix deals with navigation?
For example in our browser SDK, we use `window.location.href`, and in Next.js SDK we use `response.redirect(url)`.

packages/remix/src/infrastructure/logto/create-storage.spec.ts
```diff
@@ -0,0 +1,54 @@
...
+
+    expect(session.data.idToken).toBe('a');
+    expect(session.data.accessToken).toBeUndefined();
+  });
+
+  it('can get items', async () => {
+    const session = createSession();
+    const storage = createStorage(session);
+
+    await storage.setItem('idToken', 'a');
```
charIeszhao, line 45: nit: This one seems unnecessary now.

packages/remix/src/infrastructure/logto/handle-sign-in.ts
```diff
@@ -0,0 +1,34 @@
...
+  async (request: HandleSignInRequest): Promise<HandleSignInResponse> => {
+    const { storage, createClient } = deps;
+
+    // eslint-disable-next-line @silverhand/fp/no-let
+    let navigateToUrl = '/';
+
+    const client = createClient((url) => {
+      // eslint-disable-next-line @silverhand/fp/no-mutation
+      navigateToUrl = url;
+    });
```
charIeszhao, line 26: Hmmm, I think I get it now. But can we make `makeLogtoClient` a class, and store the navigateUrl in the class as a private property?
In this way, we don't need to create the variable here with `let` and also don't have to pass an empty `navigate` function when initiatiate the `LogtoClient` in [here](https://github.com/logto-io/js/pull/417/files#r996918007)?

charIeszhao: You can refer to our [Next.js SDK](https://github.com/logto-io/js/blob/master/packages/next/src/index.ts#L15) and let me know if that makes sense to you.

packages/remix/src/infrastructure/logto/create-client.ts
akoenig: It handles it similar to how Next.js handles redirects. You can find an example in this controller [HandleSignInCallbackController.ts#L51](https://github.com/openformation/js/blob/remix-logto/packages/remix/src/useCases/handleSignInCallback/HandleSignInCallbackController.ts#L51)

wangsijie (approved):

packages/remix/src/infrastructure/logto/create-client.ts
charIeszhao: Well, it doesn't look like what I original suggested. I was trying to make this `create-client.ts` a class instead of the two handlers. But the current design will also do the trick.
So I'm good with it for now. If we come up with a better idea in the future, we'll definitely update and merge back to your repo.

charIeszhao (approved):

PR #423, opened 2022-10-25
fix(next-sample): fix undefined in getServerSideProps

wangsijie:
## Summary

fix "undefined" issue in getServerSideProps

## Testing

Local tested


charIeszhao (approved):
LGTM

PR #425, opened 2022-10-31
fix(client): remove access token from storage after sign-out

charIeszhao:
## Summary

Fixed a bug that `accessToken` is still in local storage after successfully signing out.
Updated test case to reflect this change.

## Testing

- [x] Passed UTs


darcyYe (approved):

darcyYe:
maybe we should also check other SDKs to see whether this bug exists.

PR #426, opened 2022-10-31
refactor(js,core): decode resource dynamically from access token on saving tokens

charIeszhao:
## Summary

* Dynamically decode resource indicator from access token when persisting it in storage
* Use `decodeJwt` from `jose` library to decode claims from id token and access token

## Testing

- [x] Passed all unit tests

### DIFF
--- a/packages/client/src/errors.ts
+++ b/packages/client/src/errors.ts
@@ -7,9 +7,9 @@ const logtoClientErrorCodes = Object.freeze({
     not_found: 'Sign-in session not found.',
   },
   not_authenticated: 'Not authenticated.',
-  get_access_token_by_refresh_token_failed: 'Failed to get access token by refresh token.',
-  fetch_user_info_failed: 'Unable to fetch user info. The access token may be invalid.',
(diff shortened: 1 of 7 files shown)


wangsijie (commented):
What can we benefit from this change?

charIeszhao:
> What can we benefit from this change?

There is a case that the resource indicator you pass on `getAccessToken` could be different from the audience that encoded in the accessToken.

With this change, we can guarantee the resource we use in the access token map is the same resource that access token is issued with.

wangsijie:
I got it, thanks @charIeszhao

charIeszhao:
> I got it, thanks @charIeszhao

Hmmm, after you asking me, I re-think about this and now I feel it might not be the right thing to do. Let's put this on hold for now.

gao-sun:
should we close this? @charIeszhao

charIeszhao:
> should we close this? @charIeszhao 

Yeah, I think we should, as per our discussion this is not needed anymore.

PR #432, opened 2022-11-22
refactor(client): remove refresh token concurrent request protection

wangsijie:
## Summary

remove refresh token concurrent request protection. For now, we protect it on the server side.

reference: https://github.com/logto-io/kotlin/pull/203

## Testing

UTs


xiaoyijun (approved):
LGTM

PR #440, opened 2023-02-03
fix: remove core-kit dependency and update `engines`

gao-sun:
## Summary

- remove `@logto/core-kit` and manually add constants to avoid ESM issue
- upgrade parcel and supertest

## Testing

- CI
- local tested react/vue sample works as expected


gao-sun:
@logto-io/eng guys i'll force merge this since it affects the latest version. please take a look once get back to work

PR #442, opened 2023-02-09
fix: remove deperecated role names

wangsijie:
## Summary

Remove all `role_names` since it's deprecated.

## Testing

UTs

### DIFF
--- a/packages/js/src/utils/id-token.test.ts
+++ b/packages/js/src/utils/id-token.test.ts
@@ -213,7 +213,7 @@ describe('verifyIdToken', () => {
 describe('decodeIdToken', () => {
   test('decoding valid JWT should return claims', async () => {
     const { privateKey } = await generateKeyPair('RS256');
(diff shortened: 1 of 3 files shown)


packages/js/src/utils/id-token.test.ts
```diff
@@ -229,23 +229,9 @@ describe('decodeIdToken', () => {
...
-    const jwt = await new SignJWT({ name: null, avatar: undefined, role_names: [123] })
-      .setProtectedHeader({ alg: 'RS256' })
-      .setIssuer('foo')
-      .setSubject('bar')
-      .setAudience('qux')
-      .setExpirationTime(2000)
-      .setIssuedAt(1000)
-      .sign(privateKey);
-    expect(() => decodeIdToken(jwt)).toThrowError(TypeError);
-  });
```
charIeszhao, line 247: We should keep this test case, but exchange the wront typed `role_names` with something else

charIeszhao (approved):

PR #447, opened 2023-02-27
chore: add nullable to UserInfoResponse type properties

charIeszhao:
## Summary

Fix `UserInfoResponse` type in JS SDKs. Make most of the fields `Nullable`, which reflects the real field type in DB

## Testing

- [x] Locally tested

## Checklist

- [ ] `.changset-staged`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/js/src/core/user-info.ts
+++ b/packages/js/src/core/user-info.ts
@@ -1,3 +1,5 @@
+import { Nullable } from '@silverhand/essentials';
+
 import { Requester } from '../types';
 
 type Identity = {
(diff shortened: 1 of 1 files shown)


gao-sun (approved):

packages/js/src/core/user-info.ts
```diff
@@ -7,12 +9,12 @@ type Identity = {
...
+  name?: Nullable<string>;
+  username?: Nullable<string>;
+  picture?: Nullable<string>;
+  email?: Nullable<string>;
   email_verified?: boolean;
-  phone_number?: string;
+  phone_number?: Nullable<string>;
   phone_number_verified?: boolean;
   custom_data?: unknown;
   identities?: Record<string, Identity>;
```
gao-sun, line 20: just double check these two fields. if they won't be `null` maybe we can add some comments?

charIeszhao: I checked DB and it seems these jsonb data will have `{}` as default value. Comments added, ptal

PR #477, opened 2023-04-24
fix: typings

gao-sun:
## Summary

fixing package typings and imports by:

- add `"type": "module"` in pacakge.json to claim ESM
- compile CJS and set `main` `require` to the CJS file for compatibility
- update import path per module resolution update (Node ESM needs `.js`)
- upgrade essentials to make the dependency work with CJS
- update jest config to ESM (note the test is still running in CJS)

## Testing

tested in core. all packages build ok

## Checklist


packages/browser/src/index.ts
```diff
@@ -27,7 +27,7 @@ const navigate = (url: string) => {
   window.location.assign(url);
 };
 
-export default class LogtoClient extends BaseClient {
+export class LogtoClient extends BaseClient {
```
charIeszhao, line 30: Should we still keep this `export`, since we have `export default` in the end?

charIeszhao (approved):

simeng-li (approved):
LGTM

wangsijie (approved):

charIeszhao (approved):
Amazing!

PR #486, opened 2023-04-27
feat: next edge runtime full support

wangsijie:
## Summary

Previously, the edge runtime support for NextJs is only for "get auth info" (`withLogtoApiRoute`). Now all functions are available. That is a preparation for the upcoming Next 13 support.

The main differences in edge runtime:

1. some of (most of) the NodeJS API are not supported, such as `crypto`, it is more like the browser runtime with standard Web API.
2. there are no existing "response", it is usually created in the handler.

There are 2 main changes:
1. Add package `@logto/node/edge`
3. Rewrite `@logto/next/edge`

## Testing

Pass all tests, and local tested with edge runtime simulator.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

- [ ] This PR is not applicable for the checklist


packages/node/edge/index.ts
```diff
@@ -0,0 +1,36 @@
...
+      ...adapter,
+      requester: createRequester(
+        config.appSecret
+          ? async (...args: Parameters<typeof fetch>) => {
+              const [input, init] = args;
+
+              return fetch(input, {
+                ...init,
+                headers: {
+                  Authorization: `basic ${Buffer.from(
```
gao-sun, line 20:
```suggestion
                  Authorization: `Basic ${Buffer.from(
```

packages/node/edge/index.ts
```diff
@@ -0,0 +1,36 @@
...
+                    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
```
gao-sun, line 21: looks weird why this rule is triggered

packages/node/edge/index.ts
```diff
@@ -0,0 +1,36 @@
+import type { LogtoConfig, ClientAdapter } from '@logto/client';
+import { createRequester } from '@logto/client';
+
+import BaseClient from '../src/client.js';
+
+import { generateCodeChallenge, generateCodeVerifier, generateState } from './generators.js';
+
+export default class LogtoClient extends BaseClient {
```
gao-sun, line 8: should we add some comments to say this client is for NextJS edge runtime? also since it's exclusively for NextJS (if i understand correctly), shall we just put this into `@logto/next`?

wangsijie: This is not exclusively for NextJS, in the future we may support other edge runtime (Cloudflare workers). Comments added.

packages/node/edge/index.ts
charIeszhao: Maybe because it's nested

packages/node/src/client.ts
```diff
@@ -0,0 +1,67 @@
...
+        isAuthenticated,
+        claims: await this.getIdTokenClaims(),
+        userInfo: conditional(fetchUserInfo && (await this.fetchUserInfo())),
+        accessToken,
+      };
+    } catch {
+      return {
+        isAuthenticated: false,
+      };
+    }
```
charIeszhao, line 65: The error will be omitted in this way. Is this expected behavior?
It would be better to remove the `try catch` clause and just throw the error to the app, so that the developer knows what goes wrong

wangsijie: Well, that is an old logic, take a look

.changeset/fifty-rocks-camp.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/next-sample": minor
+"@logto/next": minor
+"@logto/node": minor
```
charIeszhao, line 4: Should we mark these as a major change? Will it break things if they use `^` with the version number in package.json?

wangsijie: The modifications will not impact the utilization, as compatibility is maintained.

packages/node/edge/index.ts
gao-sun: cool

gao-sun: did you forget to push? cannot find the comments

gao-sun (approved):
remember update the comment to node client

packages/node/edge/index.ts
wangsijie: updated

PR #496, opened 2023-05-17
fix: skip access token decode for opaque token

wangsijie:
## Summary

skip access token decode for opaque token, opaque token has "empty claims", should not throw an error.

## Testing

Local tested with express sample

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests
- [ ] docs

OR

- [ ] This PR is not applicable for the checklist

### DIFF
--- a/packages/js/src/utils/access-token.ts
+++ b/packages/js/src/utils/access-token.ts
@@ -3,7 +3,6 @@
 import { urlSafeBase64 } from '@silverhand/essentials';
 
 import { isArbitraryObject } from './arbitrary-object.js';
-import { LogtoError } from './errors.js';
 
 // Claims may vary, based on the custom OIDC config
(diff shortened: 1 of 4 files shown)


packages/js/src/utils/access-token.ts
```diff
@@ -47,7 +46,8 @@ export const decodeAccessToken = (accessToken: string): AccessTokenClaims => {
   const { 1: encodedPayload } = accessToken.split('.');
 
   if (!encodedPayload) {
-    throw new LogtoError('access_token.invalid_token');
+    // Opaque access token
```
charIeszhao, line 49: nit: We shouldn't assume it's an opaque token. You can say `non-JWT format token string` maybe?

wangsijie: updated

PR #500, opened 2023-05-23
fix: add typeVersions field in package.json of Next SDK (#499)

### LINKED ISSUE #499: bug: Cannot find module '@logto/next/edge' or its corresponding type declarations.ts(2307)
### Describe the bug
Cannot find module '@logto/next/edge' or its corresponding type declarations.ts(2307)

### Expected behavior

Importing modules without errors.

### How to reproduce?

Use "next-app-dir-sample" to change the version of "@logto/next" in packages.json dependencies to the latest version (2.1.0) published on npm. This will reproduce.

It works after import, but it does not pass the "Linting and checking validity of types"

When I moved "node_modules/@logto/next/lib/edge" to "node_modules/@logto/next/edge",  I was able to import it successfully.
I think this may be a problem caused by not export correctly?

### Context

[next-app-dir-sample](https://github.com/logto-io/js/tree/c5773294f8ab09fdbf9d23fbbed73bca15fbbb38/packages/next-app-dir-sample)

- [ ] Logto Cloud
- [x] Self-hosted, Logto version = 
  - [ ] Container (Docker image)
  - [x] Raw Node.js

### Screenshots

Snipaste_2023-05-23_17-54-35
Snipaste_2023-05-23_17-57-38

eluyuy:
## Summary

fix Cannot find module '@logto/next/edge' or its corresponding type declarations.ts (2307) (fixes #499)

Set `typesVersions` to ensure that the corresponding imported type declaration file is used correctly

ref:
https://github.com/vercel/satori/blob/341bfab3326da507accc550d0197ffed90d8e3ac/package.json
https://www.typescriptlang.org/docs/handbook/declaration-files/publishing.html#version-selection-with-typesversions

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] docs

OR

### DIFF
--- a/packages/next/package.json
+++ b/packages/next/package.json
@@ -5,6 +5,16 @@
   "main": "./lib/src/index.cjs",
   "module": "./lib/src/index.js",
   "types": "./lib/src/index.d.ts",
+  "typesVersions": {
+    "*": {
+      "*": [
+        "./lib/src/index.d.ts"
+      ],
+      "edge": [
+        "./lib/edge/index.d.ts"
+      ]
+    }
+  },
   "exports": {
     ".": {
       "require": "./lib/src/index.cjs",


wangsijie (approved):

wangsijie:
It seems like it's for compatibility with older versions of TypeScript?

eluyuy:
> 
Yes, this is the standard usage method.

However, after referring to other repositories, it is found that this can also be used to ensure the correct usage of corresponding type declaration files when importing modules from subdirectories within the Package.

Although the purpose may not be entirely standard... it is more like a kind of trick, perhaps?

packages/next/package.json
```diff
@@ -73,4 +83,4 @@
   "publishConfig": {
     "access": "public"
   }
-}
+}
```
charIeszhao, line 86:
```suggestion
}

```

charIeszhao (approved):

PR #520, opened 2023-07-05
fix: unexpected error when handling sign-in callback in React and Vue SDKs

### LINKED ISSUE #508: Bug: Incorrect Return Value from isSignInRedirected in watchEffect
https://github.com/logto-io/js/blob/4f23bf9c2693bd2420c9df22345ef44bd00907a0/packages/vue/src/index.ts#L137-L143

The code provided in the `watchEffect` block has a bug that causes the `isSignInRedirected` function to always return true, regardless of its actual logic. This issue arises because the function `isSignInRedirected` is an async function and is called without awaiting its execution.

### LINKED ISSUE #516: bug: When a page user logs in, an error will occur on the callview page, an infinite loop of printing error messages.
Using logo1.5 and @ logo/vue 2.0.1 versions,
When a user logs in, there is a small probability that an error will occur on the callview page. When checking the control, it is found that there is an infinite loop of printing error messages.
The callback page uses the official recommended writing method

使用logto1.5版本和@logto/vue 2.0.1版本,
用户登录时，小概率会出现在callview页面出错的情况，查看控制发现无限循环打印错误信息。也不给跳出的机会.
callbak page 使用的是官方推荐写法.

这个是引起错误的url
> http://127.0.0.1:5173/callback?error=invalid_target&error_description=resource%20indicator%20must%20be%20an%20absolute%20URI&state=RA8oJLat9V6-TO61udded36nYCOLu3Z0caYFYuok8yAZRKEtw-FL-q5MwvZRCDQxIbxuAm5zUjL_yLCvILeAxg&iss=https%3A%2F%2Faccount.kcpop.plus%2Foidc

image

charIeszhao:
## Summary

Fix potential failure when handling sign-in callback in React and Vue SDKs. Should fix #508 and may also fix #516 

## Testing

UT

## Checklist

### DIFF
--- a/packages/react/src/hooks/index.ts
+++ b/packages/react/src/hooks/index.ts
@@ -82,11 +82,19 @@ const useHandleSignInCallback = (callback?: () => void) => {
   );
 
   useEffect(() => {
-    const currentPageUrl = window.location.href;
-
-    if (!isAuthenticated && logtoClient?.isSignInRedirected(currentPageUrl)) {
-      void handleSignInCallback(currentPageUrl);
+    if (!logtoClient) {
+      return;
     }
+
+    (async () => {
+      const currentPageUrl = window.location.href;
+      const isAuthenticated = await logtoClient.isAuthenticated();
+      const isRedirected = await logtoClient.isSignInRedirected(currentPageUrl);
+
+      if (!isAuthenticated && isRedirected) {
+        void handleSignInCallback(currentPageUrl);
+      }
+    })();
   }, [handleSignInCallback, isAuthenticated, logtoClient]);
 
   return {
--- a/.changeset/curvy-ways-confess.md
+++ b/.changeset/curvy-ways-confess.md
@@ -1,0 +1,6 @@
+---
+"@logto/react": patch
+"@logto/vue": patch
+---
+
+Fix potential infinite loop when handling sign-in callback in React and Vue SDKs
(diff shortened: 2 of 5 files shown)


packages/react/src/hooks/index.ts
```diff
@@ -82,11 +82,22 @@ const useHandleSignInCallback = (callback?: () => void) => {
...
-      void handleSignInCallback(currentPageUrl);
+    if (!logtoClient) {
+      return;
     }
+
+    (async () => {
+      const currentPageUrl = window.location.href;
+      const isAuthenticated = await logtoClient.isAuthenticated();
+      const isRedirected = await logtoClient.isSignInRedirected(currentPageUrl);
+      console.log('########################## isAuthenticated:', isAuthenticated);
```
wangsijie, line 93: Forget to remove?

wangsijie (approved):

packages/react/src/hooks/index.ts
charIeszhao: Ah, thanks

pnpm-lock.yaml
```diff
@@ -1,4 +1,4 @@
-lockfileVersion: '6.0'
+lockfileVersion: '6.1'
```
gao-sun, line 1: should be 6.0, upgrade your local pnpm

gao-sun (approved):

pnpm-lock.yaml
charIeszhao: Ah, ok

PR #522, opened 2023-07-06
fix(client): use timestamp before issue token request to ensure expiresAt is smaller than token exp

### LINKED ISSUE #518: bug: access token expiration mismatch
### Describe the bug
The `getAccessToken()` function sometimes returns access tokens that have already expired. I found two issues with that function: 

1. Getting an access token by using `fetchTokenByRefreshToken()` seems to sometimes return access tokens that are already expired. E.g. the `exp` token inside the JWT is less than the current time. I'm not sure how that happens but the issue is most probably on the server side.
2. The second issue is using an arbitrary value for the `expiresAt` property inside the `accessTokenMap`. The way it's done right now means that there'll always be a mismatch between the real JWT expiry and the one present in the `accessTokenMap`. The `expiresAt` property should either be returned from the server, by using the same expiry that's used in the JWT, or by decoding the JWT on the client side and calculating the `expiresAt` value by using the JWT's `exp` value.

These two issues combined make it impossible to refresh the access token properly. The `accessTokenMap.expireAt` value is being used to determine whether the access token has expired and whether we need to refresh it using the refresh token. And in some cases that value is valid, meaning that the access token hasn't expired, but the JWT's `exp` value says otherwise and is being refused by the server, leading to 401.

This is a critical issue and needs to be resolved ASAP IMO. We're currently migrating from a different provider and this is a huge blocker for us and I suppose everyone else.
 
### Expected behavior
`accessTokenMap.expireAt` should always be derived from the JWT's `exp` value.

### How to reproduce?
Reduce the access token lifetime so you can force access token refresh frequently.

### Context
- [ ] Logto Cloud
- [X ] Self-hosted, Logto version = 1.5, @logto/browser = 2.1.0
  - [ X] Container (Docker image)
  - [ ] Raw Node.js

### Screenshots
CleanShot 2023-07-04 at 23 11 08@2x

charIeszhao:
## Summary

Calculate the `expiresAt` by `requestedAt + expiresIn`, which ensures the calculated result is always smaller than the actual `exp` property in token claims.

```
expiresAt = requestedAt + actual request time + expiresIn
```

Should fix #518 

## Testing

N/A

## Checklist

- [x] `.changeset`

### DIFF
--- a/packages/client/src/index.ts
+++ b/packages/client/src/index.ts
@@ -1,5 +1,4 @@
 import {
-  type CodeTokenResponse,
   type IdTokenClaims,
   type UserInfoResponse,
   type InteractionMode,
@@ -170,8 +169,7 @@ export default class LogtoClient {
   }
 
   async handleSignInCallback(callbackUri: string) {
-    const { logtoConfig, adapter } = this;
-    const { requester } = adapter;
+    const { requester } = this.adapter;
     const signInSession = await this.getSignInSession();
 
     if (!signInSession) {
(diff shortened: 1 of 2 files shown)


ikupenov (changes requested):
Cross-posting from https://github.com/logto-io/js/issues/518.

~I don't think the `expiresAt` value should be calculated by using `Now + expiresIn - shiftedTime` but rather `issueAt` (iat claim) + `expiresIn` (or any other client-specific value that's less than `expiresAt`) - `shiftedTime` (maybe not needed). Correct me if I don't understand this right.~

~To summarize, I don't think it's a good idea to use `Date.now()` when you can use either the `exp` or `iat` claim to get a much more consistent result across the client and the server.~

ikupenov (approved):

packages/client/src/index.ts
```diff
@@ -277,7 +288,7 @@ export default class LogtoClient {
     this.accessTokenMap.set(accessTokenKey, {
       token: accessToken,
       scope,
-      expiresAt: Math.round(Date.now() / 1000) + expiresIn,
+      expiresAt: requestedAt + expiresIn,
```
gao-sun, line 291: better to add some comments to the type definition of this field to clarify it's an estimate?

charIeszhao: OK

wangsijie (approved):

ikupenov (approved):

PR #545, opened 2023-09-09
refactor(next): remove iron session

wangsijie:
## Summary

Remove "iron session" dependency in Next.js SDK, replaced with our own implementation of cookie-based session.

The reason is that, "iron session" depends on `Request` and `Response`, which is not available in Next.js's Server Actions. Next, we'll add support for Server Actions.

## Testing

Pass existing tests, and tested locally with real Logto endpoint.

## Checklist

- [ ] `.changeset`
- [x] unit tests
- [x] integration tests
- [ ] docs

### DIFF
--- a/packages/next/edge/index.ts
+++ b/packages/next/edge/index.ts
@@ -1,9 +1,10 @@
+import { RequestCookies, ResponseCookies } from '@edge-runtime/cookies';
 import { type GetContextParameters, type InteractionMode } from '@logto/node';
 import NodeClient from '@logto/node/edge';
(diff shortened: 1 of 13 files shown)


charIeszhao (approved):

charIeszhao:
Do we need to draft a changeset?
Do we need to update next sample project?

wangsijie:
@charIeszhao I think the answer is no, because this is a refactor, the usage stays the same.

charIeszhao:
> @charIeszhao I think the answer is no, because this is a refactor, the usage stays the same.

OK, I'm good.

PR #562, opened 2023-09-29
feat(next): add server actions support

wangsijie:
## Summary

Add new sub-package `@logto/next/server-actions`, now Logto's Next SDK supports the new feature [Server Actions](https://nextjs.org/docs/app/api-reference/functions/server-actions).

A new example package also added.

## Testing

Local tested with the example project.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests
- [x] docs (will write a blog post later)

### DIFF
--- a/packages/next/server-actions/index.ts
+++ b/packages/next/server-actions/index.ts
@@ -1,0 +1,106 @@
+'use server';
+
+import { type GetContextParameters, type InteractionMode } from '@logto/node';
+import NodeClient from '@logto/node/edge';
+
+import BaseClient from '../src/client';
+import { createSession } from '../src/session';
+import type { LogtoNextConfig } from '../src/types.js';
+
+export type { LogtoContext, InteractionMode } from '@logto/node';
+
+export default class LogtoClient extends BaseClient {
+  constructor(config: LogtoNextConfig) {
+    super(config, {
+      NodeClient,
+    });
+  }
+
+  /**
+   * Init sign in and return the url to redirect to Logto
+   *
+   * @param cookie the raw cookie string
+   * @param redirectUri the uri (callbackUri) to redirect to after sign in
+   * @param interactionMode OIDC interaction mode
+   * @returns the url to redirect to and new cookie if any
+   */
+  async handleSignIn(
+    cookie: string,
+    redirectUri: string,
+    interactionMode?: InteractionMode
+  ): Promise<{ url: string; newCookie?: string }> {
+    const { nodeClient, session } = await this.createNodeClientFromHeaders(cookie);
+    await nodeClient.signIn(redirectUri, interactionMode);
+
+    if (!this.navigateUrl) {
+      // Not expected to happen
+      throw new Error('navigateUrl is not set');
+    }
+
+    return {
+      url: this.navigateUrl,
+      newCookie: await session.getValues?.(),
+    };
+  }
+
+  /**
+   * Init sign out and return the url to redirect to Logto
+   *
+   * @param cookie the raw cookie string
+   * @param redirectUri the uri (postSignOutUri) to redirect to after sign out
+   * @returns the url to redirect to
+   */
+  async handleSignOut(cookie: string, redirectUri = this.config.baseUrl): Promise<string> {
+    const { nodeClient } = await this.createNodeClientFromHeaders(cookie);
+    await nodeClient.signOut(redirectUri);
+
+    if (!this.navigateUrl) {
+      // Not expected to happen
+      throw new Error('navigateUrl is not set');
+    }
+
+    return this.navigateUrl;
+  }
+
+  /**
+   * Handle sign in callback from Logto
+   *
+   * @param cookie the raw cookie string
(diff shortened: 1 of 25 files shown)


packages/next-server-actions-sample/.gitignore
gao-sun, line 1: no need for this file?

packages/next-server-actions-sample/app/page.module.css
gao-sun, line 1: no need for this file?

packages/next/server-actions/index.ts
```diff
@@ -0,0 +1,80 @@
...
+export type { LogtoContext, InteractionMode } from '@logto/node';
+
+export default class LogtoClient extends BaseClient {
+  constructor(config: LogtoNextConfig) {
+    super(config, {
+      NodeClient,
+    });
+  }
+
+  handleSignIn = async (
```
gao-sun, line 19: better to use the traditional definition since we cannot bind `this` for arrow functions? this may limit some use cases

packages/next/server-actions/index.ts
gao-sun, line 1: shall we add some unit tests for this class?

packages/next/src/session.ts
```diff
@@ -104,15 +104,16 @@ type SessionConfigs = {
...
-  setCookie: (value: string) => void
+  setCookie?: (value: string) => void
 ): Promise<Session> => {
   const data = await unwrapSession(cookie, secret, crypto);
 
   const session: Session = {
     ...data,
     save: async () => {
-      setCookie(await wrapSession(session, secret, crypto));
+      setCookie?.(await wrapSession(session, secret, crypto));
```
gao-sun, line 114: use `session.getValues()`?

packages/next/server-actions/index.ts
```diff
@@ -0,0 +1,80 @@
...
+  };
+
+  private async createNodeClientFromHeaders(cookie?: string) {
+    const session = await createSession(
+      {
+        secret: this.config.cookieSecret,
+        crypto,
+      },
+      cookie ?? ''
+    );
```
gao-sun, line 74: why cookie and setCookie are optional in this case? does server actions have some magic here?

packages/next/server-actions/index.ts
gao-sun, line 19: plus, detailed JSDoc/TSDoc comments will be appreciated especially for public methods

packages/next/server-actions/index.ts
wangsijie: `cookie` is required, fixed. In server actions, it is unable to set cookies inside the SDK (can not set "use server" in library), so we use "getCookies" and then set cookies outside the sdk.

packages/next/server-actions/index.ts
wangsijie: updated

packages/next/server-actions/index.ts
wangsijie: updated

packages/next/server-actions/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    if (!this.navigateUrl) {
+      // Not expected to happen
+      throw new Error('navigateUrl is not set');
+    }
+
+    return this.navigateUrl;
+  }
+
+  /**
+   * Handle sign in callback from Logto
```
gao-sun, line 66:
```suggestion
   * Handle sign-in callback from Logto.
```

packages/next/server-actions/index.ts
```diff
@@ -0,0 +1,106 @@
...
+    }
+
+    return {
+      url: this.navigateUrl,
+      newCookie: await session.getValues?.(),
+    };
+  }
+
+  /**
+   * Init sign out and return the url to redirect to Logto
```
gao-sun, line 47:
```suggestion
   * Init sign-out and return the url to redirect to Logto.
```

packages/next/server-actions/index.ts
```diff
@@ -0,0 +1,106 @@
...
+
+export default class LogtoClient extends BaseClient {
+  constructor(config: LogtoNextConfig) {
+    super(config, {
+      NodeClient,
+    });
+  }
+
+  /**
+   * Init sign in and return the url to redirect to Logto
```
gao-sun, line 20:
```suggestion
   * Init sign-in and return the url to redirect to Logto.
```

packages/next/server-actions/index.ts
```diff
@@ -0,0 +1,106 @@
...
+   */
+  async handleSignInCallback(cookie: string, callbackUrl: string): Promise<string | undefined> {
+    const { nodeClient, session } = await this.createNodeClientFromHeaders(cookie);
+
+    await nodeClient.handleSignInCallback(callbackUrl);
+    return session.getValues?.();
+  }
+
+  /**
+   * Get Logto context from cookies
```
gao-sun, line 80:
```suggestion
   * Get Logto context from cookies.
```

gao-sun (approved):

PR #586, opened 2023-11-30
refactor(react)!: refactor LogtoContextProps and LogtoContext

gao-sun:
## Summary

- remove `useLoadingState`, leave the logic inside the context since they are generic.
- export more useful `isLoading` and `setIsLoading` from Logto context.
- export `LogtoContext` from the package

## Testing

tested in Logto demo app. works as expected.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/wicked-peas-destroy.md
+++ b/.changeset/wicked-peas-destroy.md
@@ -1,0 +1,11 @@
+---
+"@logto/react": major
+---
+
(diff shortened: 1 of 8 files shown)


charIeszhao (approved):
I think we can move on and upgraded it in console once we release the next version.

PR #591, opened 2023-12-05
refactor(js,client,node): use btoa to replace buffer

wangsijie:
## Summary

Use `btoa` function in NodeJS environment. Although this is deprecated, but it still works. 

`Buffer` is not available in some edge runtime, for example, in Cloudflare Pages, it should be imported through:

```
import { Buffer } from 'node:buffer";
```

But this import will fail in browser environment.

Also updated the package `@silverhand/essentials`, see https://github.com/silverhand-io/essentials/pull/58

The new version of `@silverhand/essentials` causes some issues with types, to bypass these, use explicit type annotations.

## Testing

Pass existing tests.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/node/edge/index.ts
+++ b/packages/node/edge/index.ts
@@ -17,14 +17,13 @@ export default class LogtoClient extends BaseClient {
           ? async (...args: Parameters<typeof fetch>) => {
               const [input, init] = args;
 
+              // Encode to base64 using btoa
+              const base64Credentials = btoa(`${config.appId}:${config.appSecret ?? ''}`);
+
               return fetch(input, {
                 ...init,
                 headers: {
-                  Authorization: `Basic ${Buffer.from(
-                    // eslint-disable-next-line @typescript-eslint/restrict-template-expressions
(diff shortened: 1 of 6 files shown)


charIeszhao (approved):

packages/node/edge/index.ts
```diff
@@ -17,14 +17,13 @@ export default class LogtoClient extends BaseClient {
           ? async (...args: Parameters<typeof fetch>) => {
               const [input, init] = args;
 
+              // Encode to base64 using btoa
+              const base64Credentials = btoa(`${config.appId}:${config.appSecret ?? ''}`);
```
charIeszhao, line 21: nit: use `window.btoa` to workaround the deprecated mark?

charIeszhao (approved):

packages/node/edge/index.ts
wangsijie: in Node.js, `window` is not available, however, `globalThis` is OK

PR #635, opened 2024-02-21
refactor: implement `CookieStorage` for node sdk

gao-sun:
## Summary

- implement reusable `CookieStorage` for Node environment (extracted from `@logto/sveltekit`)
- remove `node-fetch` from `@logto/node` (Node comes with native fetch from 17)
- minor package metadata updates
- implement a simple Promise queue as `p-queue` is ESM only but we still target CJS

## Testing

unit tests + local sign-in/out

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/node/src/utils/promise-queue.test.ts
+++ b/packages/node/src/utils/promise-queue.test.ts
@@ -1,0 +1,57 @@
+import { PromiseQueue } from './promise-queue.js';
+
+/* eslint-disable @silverhand/fp/no-mutating-methods */
+describe('PromiseQueue', () => {
+  it('should process tasks sequentially in the order they are enqueued', async () => {
+    const queue = new PromiseQueue();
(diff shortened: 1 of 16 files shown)


packages/node/src/utils/promise-queue.test.ts
wangsijie, line 1: Want to make sure this file is not checked by accident, this change is not included in PR's title or description

wangsijie (approved):

packages/node/src/utils/promise-queue.test.ts
gao-sun: my bad. i was using `p-queue` but it's an ESM-only package. so i removed it and added this

gao-sun: updated pull description

PR #657, opened 2024-03-11
feat(js): add secureRoutes support to buildAngularAuthConfig util

charIeszhao:
## Summary

Add support to `secureRoutes` config in `buildAngularAuthConfig` util, so that `angular-auth-oidc-client` will automatically add access tokens to request header for these routes.

Feature requested by community:
https://discord.com/channels/965845662535147551/1215176080680550461

## Testing

We can do this now:
image

## Checklist

- [x] `.changeset`

~~- [ ] unit tests~~
~~- [ ] integration tests~~
~~- [ ] necessary TSDoc comments~~

### DIFF
--- a/packages/js/src/utils/angular.ts
+++ b/packages/js/src/utils/angular.ts
@@ -5,7 +5,7 @@ import { Prompt } from '../consts/index.js';
 import { withDefaultScopes } from './scopes.js';
 
 /** The Logto configuration object for Angular apps. */
-export type LogtoAngularConfig = {
+export type LogtoAngularConfig = OpenIdConfiguration & {
   /**
    * The endpoint for the Logto server, you can get it from the integration guide
    * or the team settings page of the Logto Console.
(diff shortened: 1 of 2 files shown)


packages/js/src/utils/angular.ts
```diff
@@ -46,6 +46,11 @@ export type LogtoAngularConfig = {
    * @default Prompt.Consent
    */
   prompt?: Prompt | Prompt[];
+  /**
+   * The routes protected by the `angular-auth-oidc-client` library.
+   * The access token will be added automatically to the `Authorization` header for these routes.
+   */
+  secureRoutes?: string[];
```
gao-sun, line 53: This function was intended to transpile the options into a "Logto" way. But this one is the option from the Angular library. We can use object spreading to merge configs instead of adding a new key for this function?

charIeszhao: Yeah, that makes sense

charIeszhao: Code updated to allow any `OpenIdConfiguration` config to be indicated in `buildAngularAuthConfig` util, which would be spreaded and returned as-is.

charIeszhao:
After discussion, we've decided not to mix Logto configs with angular-oidc-auth-client configs, since some of the params are conflicting and might cause further confusion. Closing the ticket...

PR #661, opened 2024-03-21
feat(next): add error handler to withLogtoApiRoute

wangsijie:
## Summary

Add error handler to withLogtoApiRoute, this is useful when developers need to handle the auth failed case.

## Testing

Unit tests.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/spicy-bobcats-brush.md
+++ b/.changeset/spicy-bobcats-brush.md
@@ -1,0 +1,5 @@
+---
+"@logto/next": minor
+---
+
+Add error handler to Next.js SDK's withLogtoApiRoute.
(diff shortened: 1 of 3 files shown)


.changeset/spicy-bobcats-brush.md
```diff
@@ -0,0 +1,5 @@
+---
+"@logto/next": minor
```
charIeszhao, line 2: Should this be a patch release?

charIeszhao (approved):

.changeset/spicy-bobcats-brush.md
wangsijie: New api function added, so I think it is a minor release

PR #672, opened 2024-03-22
refactor: fix types

gao-sun:
## Summary

- overload signature cannot be picked up in react, this pull request fixes it.
- remove unsupported direct sign-in methods

## Testing

- added unit test
- correctly infer the type

image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/chilly-pets-walk.md
+++ b/.changeset/chilly-pets-walk.md
@@ -1,0 +1,7 @@
+---
+"@logto/react": patch
+---
+
+fix `signIn()` type
+
+Now it can correctly infer the overload signatures.
--- a/packages/react/src/hooks/index.ts
+++ b/packages/react/src/hooks/index.ts
@@ -24,11 +24,12 @@ type Logto = {
     | 'getOrganizationTokenClaims'
     | 'getIdToken'
     | 'getIdTokenClaims'
-    | 'signIn'
     | 'signOut'
     | 'fetchUserInfo'
   >
->;
+> &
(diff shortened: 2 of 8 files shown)


packages/react/src/hooks/index.ts
```diff
@@ -24,11 +24,12 @@ type Logto = {
...
     | 'getIdToken'
     | 'getIdTokenClaims'
-    | 'signIn'
     | 'signOut'
     | 'fetchUserInfo'
   >
->;
+> &
+  // Manually pick the method with overloads since TypeScript cannot infer the correct type.
+  Pick<LogtoClient, 'signIn'>;
```
charIeszhao, line 32: Just curious why would this work? It doesn't seem to do anything different. 🤔

.changeset/chilly-pets-walk.md
```diff
@@ -0,0 +1,7 @@
+---
```
charIeszhao, line 1: Would this be a breaking change? As it seems the original type infer does not work any more

packages/react/src/hooks/index.ts
gao-sun: idk, ts cannot infer via a remapping

.changeset/chilly-pets-walk.md
gao-sun: this was an experimental option and it doesn't have any effect now

charIeszhao (approved):

PR #697, opened 2024-04-16
feat(node): add access token organization id param to getContext

wangsijie:
## Summary

Add `accessTokenOrgnizationId` param to `getContext`, allowing users to set `orgniazation_id` when granting access token.

## Testing

Local tested.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/spotty-readers-rule.md
+++ b/.changeset/spotty-readers-rule.md
@@ -1,0 +1,7 @@
+---
+"@logto/node": minor
+---
+
+Add `organizationId` to `getContext` function.
+
+This will allow the users to get access token with "organization_id" claim, which supports organization API resources.
--- a/packages/node/src/client.ts
+++ b/packages/node/src/client.ts
@@ -31,6 +31,7 @@ export default class LogtoNodeBaseClient extends BaseClient {
     resource,
     fetchUserInfo,
     getOrganizationToken,
+    organizationId,
   }: GetContextParameters = {}): Promise<LogtoContext> => {
     const isAuthenticated = await this.isAuthenticated();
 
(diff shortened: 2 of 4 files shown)


packages/node/src/client.ts
```diff
@@ -31,6 +31,7 @@ export default class LogtoNodeBaseClient extends BaseClient {
     resource,
     fetchUserInfo,
     getOrganizationToken,
+    accessTokenOrganizationId,
```
charIeszhao, line 34: Would it be OK to just use `organizationId`?

packages/node/src/client.ts
```diff
@@ -44,7 +45,9 @@ export default class LogtoNodeBaseClient extends BaseClient {
 
     const { accessToken, accessTokenClaims } = getAccessToken
       ? {
-          accessToken: await trySafe(async () => this.getAccessToken(resource)),
+          accessToken: await trySafe(async () =>
+            this.getAccessToken(resource, accessTokenOrganizationId)
+          ),
```
charIeszhao, line 50: Just to confirm, do we need to update other JS SDKs?

packages/node/src/client.ts
wangsijie: To avoid confusing with "organization token".

wangsijie: Per discussion, changed to `organizationId`, we will deprecate `getOrganizationToken` later.

.changeset/spotty-readers-rule.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/node": minor
+---
+
+Add `accessTokenOrganizationId` to `getContext` function.
```
charIeszhao, line 5: Should also update the variable name in changeset

packages/node/src/index.test.ts
```diff
@@ -50,7 +50,21 @@ describe('LogtoClient', () => {
       ).resolves.toEqual({
         isAuthenticated: false,
       });
-      expect(getAccessToken).toHaveBeenCalledWith('resource');
+      expect(getAccessToken).toHaveBeenCalledWith('resource', undefined);
+    });
+
+    it('should get access token with organization_id', async () => {
```
charIeszhao, line 56: nit:
```suggestion
    it('should get access token with organization id', async () => {
```

charIeszhao (approved):

PR #720, opened 2024-05-16
fix(next): support customize redirectUri in handleSignIn

### LINKED ISSUE #701: bug:
### Describe the bug
[`client.handleSignIn`](https://github.com/logto-io/js/blob/a8e2d9386559a87282d1621fe1955163a1e300c4/packages/next/server-actions/index.ts#L24) and [`client.handleSignInCallback`](https://github.com/logto-io/js/blob/a8e2d9386559a87282d1621fe1955163a1e300c4/packages/next/server-actions/index.ts#L45) functions do not have a consistent `redirectUri` parameter.

I created my callback endpoint at the `/api/auth/callback` path, which I also specified in the cloud application config "Redirect URI", I can pass to the `signIn` a different `redirectUri` but the function to validate it `client.handleSignInCallback` do not expect to be on a different URL:
https://github.com/logto-io/js/blob/a8e2d9386559a87282d1621fe1955163a1e300c4/packages/next/server-actions/index.ts#L43-L46

### Expected behavior
The callback path should be flexible (on any configured path) or enforced to always be at `/callback`. And the `baseUrl` parameter doesn't sound like a place to include the callback url

To do it myself, the `handleSignIn` function doesn't have the option to change the callback path https://github.com/logto-io/js/blob/a8e2d9386559a87282d1621fe1955163a1e300c4/packages/next/server-actions/index.ts#L45

### How to reproduce?
1. Copy the example about `next/server-actions`
2. Move the callback route
3. Change the config to allow for a different place of this callback `http://localhost:3000/api/auth/callback` instead of `http://localhost:3000/callback`
4. You can click the sign-in button and log in
5. After the redirect you get a 500 HTTP error

### Context

- [x] Logto Cloud
- [ ] Self-hosted, Logto version = 
  - [ ] Container (Docker image)
  - [ ] Raw Node.js

### Screenshots

image

wangsijie:
fixed #701

## Summary

support custom redirectUri for handleSignIn function in Server Actions.

## Testing

Local tested.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/next/server-actions/index.ts
+++ b/packages/next/server-actions/index.ts
@@ -35,14 +35,15 @@ export const signIn = async (
  */
 export const handleSignIn = async (
   config: LogtoNextConfig,
-  searchParams: URLSearchParams
+  searchParams: URLSearchParams,
+  redirectUri?: string
 ): Promise<void> => {
   const search = searchParams.toString();
 
   const client = new LogtoClient(config);
   const newCookie = await client.handleSignInCallback(
     await getCookies(config),
-    `${config.baseUrl}/callback?${search}`
+    redirectUri ?? `${config.baseUrl}/callback?${search}`
   );
   if (newCookie) {
     await setCookies(newCookie, config);
--- a/.changeset/weak-cows-hope.md
+++ b/.changeset/weak-cows-hope.md
@@ -1,0 +1,5 @@
+---
+"@logto/next": patch
+---
+
+support custom redirectUri for handleSignIn function in Server Actions.


xiaoyijun (approved):

packages/next/server-actions/index.ts
```diff
@@ -35,14 +35,15 @@ export const signIn = async (
  */
 export const handleSignIn = async (
   config: LogtoNextConfig,
-  searchParams: URLSearchParams
+  searchParams: URLSearchParams,
+  redirectUri?: string
```
gao-sun, line 39: should be one single param? e.g. `searchParams: URLSearchParam | URL` (you can use function override for better documenting)

PR #726, opened 2024-05-24
feat(next): add error handler to handleAuthRoutes

wangsijie:
## Summary

Add a param to `handleAuthRoutes`, allowing users to handle errors.

## Testing

Existing tests should pass.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/lemon-plants-lay.md
+++ b/.changeset/lemon-plants-lay.md
@@ -1,0 +1,7 @@
+---
+"@logto/next": minor
+---
+
+add error handler to handleAuthRoutes
+
+You can now pass a custom error handler to handleAuthRoutes. This will allow you to handle errors in a way that makes sense for your application.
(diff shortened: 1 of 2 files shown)


.changeset/lemon-plants-lay.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/next": minor
+---
+
+Add error handler to handleAuthRoutes
```
charIeszhao, line 5:
```suggestion
add error handler to handleAuthRoutes
```

packages/next/src/index.ts
```diff
@@ -96,31 +96,42 @@ export default class LogtoClient extends LogtoNextBaseClient {
...
+        }
+
+        response.status(404).end();
+      } catch (error: unknown) {
+        if (onError) {
+          return onError(request, response, error);
+        }
 
-      response.status(404).end();
+        throw error;
```
charIeszhao, line 133: Should we still throw the error if it is already handled by the `onError` handler?

wangsijie: No, then the page will fallback into the service provider's default error page, and this is what `onError` is trying to fix.

charIeszhao (approved):

PR #769, opened 2024-07-15
feat(node): ignore fetch user info error in getContext

wangsijie:
## Summary

Wrap `getUserInfo` with `trySafe` in node SDK's `getContext` method. This method should not throw error if `getUserInfo` failed, it should return undefined for better development experience. If the developer want the handle the error manually, he can use `client.getUserInfo` instead.

This PR is similar to https://github.com/logto-io/js/pull/739

## Testing

Local tested and exiting tests shoudl pass.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/dry-bulldogs-travel.md
+++ b/.changeset/dry-bulldogs-travel.md
@@ -1,0 +1,9 @@
+---
+"@logto/node": minor
+---
+
+return undefined when fetch user info failed in getContext
+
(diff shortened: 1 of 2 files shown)


wangsijie:
This is a breaking change, per discussion, close for now. Open another PR when we are ready.

PR #776, opened 2024-07-27
fix: remove import from dev dependency

vstremnitzer:
## Summary
DevDependency not available in angular app

DevDependency import from **vitest** introduced in PR https://github.com/logto-io/js/pull/760 

This causes build problems in angular app as test runner is still **jest** 

## Testing

Executed all unit tests

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/js/src/core/user-info.ts
+++ b/packages/js/src/core/user-info.ts
@@ -1,3 +1,5 @@
+import { type Nullable } from '@silverhand/essentials';
+
 import { type IdTokenClaims } from '../index.js';
 import type { Requester } from '../types/index.js';
 
@@ -9,7 +11,7 @@ type Identity = {
 type OrganizationData = {
   id: string;
   name: string;
-  description?: string;
+  description: Nullable<string>;
 };
 
 export type UserInfoResponse = IdTokenClaims & {
--- a/packages/js/src/core/user-info.ts
+++ b/packages/js/src/core/user-info.ts
@@ -1,5 +1,3 @@
-import { type Nullable } from 'vitest';
-
 import { type IdTokenClaims } from '../index.js';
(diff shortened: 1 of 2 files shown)


charIeszhao:
Sorry I don't quite get the connection between your PR description and file changes.

How's the type change `Nullable -> Optional` solve your dev dependencies issue?

vstremnitzer:
Hi @charIeszhao 
native optional instead of Nullable seemed to bring the same value to the object for me. To me it seemed adding Nullable was a mistake here as it is not used in a testclass. Sorry if I am wrong.


To the dependency problem.

import { type Nullable } from 'vitest';

vitest is configured as devdependency in logtojs.

**Vitest** is not available in node_modules in default angular app as Jest runner is the default there.

https://github.com/angular/angular-cli/issues/25217

KR

packages/js/src/core/user-info.ts
```diff
@@ -11,7 +9,7 @@ type Identity = {
 type OrganizationData = {
   id: string;
   name: string;
-  description: Nullable<string>;
+  description?: string;
```
charIeszhao, line 12: Should we use `Nullable` from `@logto/essentials` instead? It sesms we were using a wrong import source

charIeszhao: @wangsijie

charIeszhao:
@vstremnitzer I just pushed a new update as it seems to be caused by a wrong import source of the type `Nullable`. So instead of changing it to `Optional`, we import it from `@silverhand/essentials`, which is included in the `dependencies`.

Can you please verify if this update works for you. Seems everything is working properly on my side.

wangsijie (approved):

PR #786, opened 2024-08-15
feat: add first screen values and identifier config for sign-in options

xiaoyijun:
## Summary

> Note: this pr will be merged once the first screen feature enhancement is released.

- We support identifier sign-in / sign-up / sso page as the first screen, so add related candidate values for the first screen param
- Add `identifiers` sign-in option config items to support generating identifier params for sign-in URL generation

## Testing

UT && Test locally
Use the following sign-in options:

image

We'll be redirect to the related identifier sign-in page with identfiers filtered:
image

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/blue-phones-work.md
+++ b/.changeset/blue-phones-work.md
@@ -1,0 +1,19 @@
+---
+"@logto/client": minor
+"@logto/js": minor
+---
+
+add support for identifier-based and single sign-on first screens and configurable identifiers
+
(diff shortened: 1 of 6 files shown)


.changeset/blue-phones-work.md
```diff
@@ -0,0 +1,17 @@
+---
+"@logto/client": minor
+"@logto/js": minor
+---
+
+add support for identifier-based ans single sign-on first screens and configurable identifiers
```
wangsijie, line 6: typo? "ans"

wangsijie (approved):

.changeset/blue-phones-work.md
xiaoyijun: updated.

PR #813, opened 2024-10-09
fix(node): remove global.crypto

wangsijie:
## Summary

The crypto parameter has been removed from the `unwrapSession` and `wrapSession` functions. With the introduction of Node.js 20, accessing crypto directly is now possible, which resolves issues related to `global is not defined` in edge runtimes.

## Testing

Local tested with Node.js and Vercel edge runtime.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/shaggy-cups-switch.md
+++ b/.changeset/shaggy-cups-switch.md
@@ -1,0 +1,8 @@
+---
+"@logto/next-sample": patch
+"@logto/node": patch
+---
+
+remove the parameter of crypto, fix global undefined error in edge runtime
+
+Remove the default `crypto` parameter in `unwrapSession` and `wrapSession`, `global.crypto` is unavailable in edge runtime, since Node.js 20, we can access `crypto` directly, this also works in edge runtime like Vercel.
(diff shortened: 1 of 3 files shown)


.changeset/shaggy-cups-switch.md
```diff
@@ -0,0 +1,8 @@
+---
+"@logto/next-sample": patch
+"@logto/node": patch
+---
+
+remove the parameter of crypto, fix global undefined error in edge runtime
+
+Remove the default `crypto` parameter in `unwrapSession` and `wrapSession`, `global.crypto` is unavailable in edge runtime, since Node.js 20, we can access `crypto` directly, this also works in edge runtime like Vercel.
```
charIeszhao, line 8: So are we having `node > 20` in the project config?

.changeset/shaggy-cups-switch.md
```diff
@@ -0,0 +1,8 @@
+---
+"@logto/next-sample": patch
+"@logto/node": patch
```
charIeszhao, line 3: Seems like a breaking change, since the node 18 will no longer be supported

charIeszhao (approved):

PR #839, opened 2024-10-29
fix(nuxt): remove default value of customRedirectBaseUrl

wangsijie:
## Summary

fix the default value of customRedirectBaseUrl

Previously, it was set to `<replace-with-custom-redirect-base-url>`, which is not a valid value, and will cause 500 error.

Now the default value is removed because it's not a required value.

## Testing

Local tested.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/nine-meals-behave.md
+++ b/.changeset/nine-meals-behave.md
@@ -1,0 +1,7 @@
+---
+"@logto/nuxt": patch
+---
+
+fix the default value of customRedirectBaseUrl
+
+Previously, it was set to '<replace-with-custom-redirect-base-url>', which is not a valid value. Now the default value is removed because it's not a required value.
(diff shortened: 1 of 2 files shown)


.changeset/nine-meals-behave.md
```diff
@@ -0,0 +1,7 @@
+---
+"@logto/nuxt": patch
+---
+
+fix the default value of customRedirectBaseUrl
+
+Previously, it was set to '<replace-with-custom-redirect-base-url>', which is not a valid value. Now the default value is removed because it's not a required value.
```
gao-sun, line 7: need a new line

gao-sun (approved):

PR #791, opened 2024-09-09
refactor(js): refactor @logto/client package exports

simeng-li:
## Summary

Refactor `@logto/client` package exports condition for react-native.
Add `react-native` package export condition.

[Enabling package exports] feature (https://reactnative.dev/blog/2023/06/21/package-exports-support#enabling-package-exports-beta) in react-native is unstable and can cause issues.

Replace the `exports` in `@logto/client` package.json with the `react-native` [condition](https://reactnative.dev/blog/2023/06/21/package-exports-support#the-new-react-native-condition).

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests
- [ ] necessary TSDoc comments

PR #866, opened 2024-12-03
feat: add package exports compatibility for react-native apps

icodesign:
## Summary

The new 0.4.0 react-native package release breaks if developers have unstable_enablePackageExports set to true, since https://github.com/logto-io/js/pull/791 uses root react-native fields in package.json.

While in office [React Native](https://reactnative.dev/blog/2023/06/21/package-exports-support#the-new-react-native-condition) and [metro](https://metrobundler.dev/docs/package-exports/#replacing-browser-and-react-native-fields) docs, enablePackageExports is going to be turned on by default in the future and the suggested way is to put react-native field under exports (instead of root).

To support both situations, we can add `react-native` field under both root and `exports`.

Discussion reference: https://github.com/logto-io/react-native/issues/20#issuecomment-2513375814

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/.changeset/chatty-dingos-confess.md
+++ b/.changeset/chatty-dingos-confess.md
@@ -1,0 +1,5 @@
+---
+"@logto/client": patch
+---
+
+fix react-native compatibility issues by re-exporting the shim
--- a/packages/client/package.json
+++ b/packages/client/package.json
@@ -7,6 +7,7 @@
   "react-native": "./lib/shim.js",
   "exports": {
     "types": "./lib/index.d.ts",
+    "react-native": "./lib/shim.js",
     "import": "./lib/index.js",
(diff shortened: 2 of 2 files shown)


simeng-li (approved):

simeng-li:
@icodesign Thanks for the PR! 🙌

@charIeszhao can we manage a new release after this PR is merged?

icodesign:
I've made a minor order change since my recent local test showed that React Native will map the first one in the condition names.

The doc says:
```
unstable_conditionNames: The set of condition names to assert when resolving conditional exports. By default, we match ['require', 'import', 'react-native'].
```

charIeszhao (approved):

PR #892, opened 2024-12-27
feat(nuxt): export event handler

wangsijie:
## Summary

Export `logtoEventHandler` so that users can use Logto in middelware like:

```ts
import { defineEventHandler, useRuntimeConfig } from '#imports';
import { logtoEventHandler } from '@logto/nuxt';

export default defineEventHandler(async (event) => {
  const config = useRuntimeConfig(event);
  await logtoEventHandler(event, config);
  console.log(event.context.logtoUser, event.context.logtoClient);
});
```

## Testing

Local tested.

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/nuxt/src/runtime/utils/handler.ts
+++ b/packages/nuxt/src/runtime/utils/handler.ts
@@ -1,0 +1,97 @@
+import LogtoClient, { CookieStorage } from '@logto/node';
+import { trySafe } from '@silverhand/essentials';
+import { type H3Event, getRequestURL, getCookie, setCookie, sendRedirect } from 'h3';
+import type { RuntimeConfig } from 'nuxt/schema';
+
+import { defaults } from './constants';
+import { type LogtoRuntimeConfig } from './types';
+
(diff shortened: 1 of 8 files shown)


packages/nuxt/src/runtime/utils/handler.ts
```diff
@@ -0,0 +1,99 @@
...
+  }
+
+  // eslint-disable-next-line @silverhand/fp/no-mutation
+  event.context.logtoClient = logto;
+  // eslint-disable-next-line @silverhand/fp/no-mutation
+  event.context.logtoUser = (await logto.isAuthenticated())
+    ? await trySafe(async () => (fetchUserInfo ? logto.fetchUserInfo() : logto.getIdTokenClaims()))
+    : undefined;
+
+  console.log('in-module');
```
charIeszhao, line 98: Remove it?

charIeszhao (approved):

PR #896, opened 2024-12-31
feat(next): external storage

wangsijie:
## Summary

This PR adds support for custom external session storage.

Add `sessionWrapper` to the config, you can implement your own session wrapper to support custom external session storage.

Take a look at the example below:

```ts
import { MemorySessionWrapper } from './storage';

export const config = {
  // ...
  sessionWrapper: new MemorySessionWrapper(),
};
```

```ts
import { randomUUID } from 'node:crypto';

import { type SessionWrapper, type SessionData } from '@logto/next';

export class MemorySessionWrapper implements SessionWrapper {
  private readonly storage = new Map<string, unknown>();

  async wrap(data: unknown, _key: string): Promise<string> {
    const sessionId = randomUUID();
    this.storage.set(sessionId, data);
    return sessionId;
  }

  async unwrap(value: string, _key: string): Promise<SessionData> {
    if (!value) {
      return {};
    }

    const data = this.storage.get(value);
    return data ?? {};
  }
}
```

You can implement your own session wrapper to support custom external session storage like Redis, Memcached, etc.

## Testing

Tested with sample projects.

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests


.changeset/neat-glasses-think.md
```diff
@@ -0,0 +1,9 @@
+---
+"@logto/browser": patch
+---
+
+support ssr for browser-based SDKs
+
+Check if the `window` is defined, if not, ignore the storage operations, avoid breaking the server side rendering.
+
+For this kind of SDKs, it doesn't make sense to be used in server side, because the auth state is usually stored in the client side which is not available in server side. So we can safely ignore the storage operations in server side.
```
charIeszhao, line 9: I'm actually a bit baffled. The title says "support" ssr for browser based SDKs, but the description says it doesn't make sense to be used in server side (which I thought is true).
Can you update the changelog to make it clear?

wangsijie: rebased, this file is gone

charIeszhao (approved):

PR #946, opened 2025-04-23
fix(js): add react-native export declarations

simeng-li:
## Summary

Add react-native specific export entry declarations to the @logto/js package. This will address the issue https://github.com/logto-io/react-native/issues/47

React Native uses a different module resolution strategy compared to Node.js. We need to ensure that the paths we are using in the exports are compatible with how React Native resolves modules. 

## Checklist

- [x] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/js/package.json
+++ b/packages/js/package.json
@@ -4,9 +4,11 @@
   "type": "module",
   "module": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "react-native": "./lib/index.js",
   "exports": {
     "types": "./lib/index.d.ts",
-    "import": "./lib/index.js"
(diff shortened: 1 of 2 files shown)


packages/js/package.json
```diff
@@ -4,9 +4,11 @@
   "type": "module",
   "module": "./lib/index.js",
   "types": "./lib/index.d.ts",
+  "react-native": "./lib/index.js",
```
charIeszhao, line 7: nit: Maybe explain why this fix works? I'm a bit clueless about this change

charIeszhao (approved):
Fix the typos and we're good to go. Approve in prior to unblock

PR #967, opened 2025-05-22
chore: fix deploy with vercel button

wangsijie:
## Summary

This PR fixes the "deploy with Vercel" button:

1. Fix package path
2. Remove optional envs
3. Use `latest` instead of `workspace`

## Testing

Tested with Vercel.

## Checklist

- [ ] `.changeset`
- [ ] unit tests
- [ ] integration tests

### DIFF
--- a/packages/next-server-actions-sample/package.json
+++ b/packages/next-server-actions-sample/package.json
@@ -9,7 +9,7 @@
     "start": "next start"
   },
   "dependencies": {
-    "@logto/next": "workspace:^",
+    "@logto/next": "^4.2.3",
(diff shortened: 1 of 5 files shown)


packages/next-server-actions-sample/package.json
```diff
@@ -9,7 +9,7 @@
     "start": "next start"
   },
   "dependencies": {
-    "@logto/next": "workspace:^",
+    "@logto/next": "latest",
```
simeng-li, line 12: Would it be better to lock the major version?

Issue #1009, opened 2025-10-28
bug: logto/client getAccessToken with a resource does not include the scopes in the token


maddijoyce:
### Describe the bug

Using logto/node, I'm authenticating my user then attempting to get a JWT access token, so I call
```
const token = await client.getAccessToken(resourceUrl);
```
The token is successfully generated and returned, however it does not have any scopes.

### Expected behavior

I believe the token should have the scopes from the options when the client is initialised, but looking at the source - https://github.com/logto-io/js/blob/c540dd667802c0372aa7891100d16bb36c758580/packages/client/src/client.ts#L449 - it does not include scopes in this call.

I'm happy to add a PR for this, just wanted to check this was in fact a bug.

### How to reproduce?

- Create a new Client from @logto/node
- Authenticate the client
- Call `getAccessToken` with a resource.

### Environment

Self-hosted (Docker image)

wangsijie:
Hi, you need to add the resource and scopes when initializing the Logto client, then use that client inside this function.

wangsijie:
@maddijoyce Yeah, that’s not intentional. It’s okay to pass an extra scope param to `fetchTokenByRefreshToken`, the backend API handles it.

250king:
Well, I have same proble about it: why it don't give reources to fetchTokenByAuthorizationCode?

<img width="1041" height="200" alt="Image" src="https://github.com/user-attachments/assets/bae25747-2a6f-4907-b2b5-030dbee85f87" />

<img width="1349" height="477" alt="Image" src="https://github.com/user-attachments/assets/ea1df62a-db49-4352-98b7-3da9a69a6f46" />

https://docs.logto.io/zh-CN/authorization/global-api-resources#authorization-code-or-refresh-token-flow

https://docs.logto.io/zh-CN/authorization/role-based-access-control#default-api-resource-behavior

It leds that I get a opaque token and it isn't included the scope I have give.

<img width="2233" height="70" alt="Image" src="https://github.com/user-attachments/assets/d146a513-5b18-4f71-9d8e-a65a8837a0df" />

<img width="778" height="366" alt="Image" src="https://github.com/user-attachments/assets/0dfddd06-3833-498c-95a3-dbc719fe11be" />

wangsijie:
[...]

PR #1013, opened 2025-11-19
fix(client): include scopes in access token request

maddijoyce:
## Summary
As per https://github.com/logto-io/js/issues/1009, when requesting an access token for a different resource, it should include the scopes from the client options.

## Testing

Updated unit tests for getAccessToken call
Ran getAccessToken call against my logto instance

## Checklist

- [X] `.changeset`
- [X] unit tests
- [X] integration tests


maddijoyce:
Reopening with signed commits

PR #1015, opened 2025-11-20
fix(client): include scopes in access token request

maddijoyce:
## Summary
As per https://github.com/logto-io/js/issues/1009, when requesting an access token for a different resource, it should include the scopes from the client options.

## Testing

Updated unit tests for getAccessToken call
Ran getAccessToken call against my logto instance

## Checklist

- [X] `.changeset`
- [X] unit tests
- [X] integration tests
- [X] necessary TSDoc comments

### DIFF
--- a/packages/client/src/client.ts
+++ b/packages/client/src/client.ts
@@ -443,7 +443,7 @@ export class StandardLogtoClient {
     }
 
     const accessTokenKey = buildAccessTokenKey(resource, organizationId);
-    const { appId: clientId } = this.logtoConfig;
+    const { appId: clientId, scopes } = this.logtoConfig;
     const { tokenEndpoint } = await this.getOidcConfig();
     const requestedAt = Math.round(Date.now() / 1000);
     const { accessToken, refreshToken, idToken, scope, expiresIn } = await fetchTokenByRefreshToken(
@@ -453,6 +453,7 @@ export class StandardLogtoClient {
         refreshToken: currentRefreshToken,
         resource,
         organizationId,
+        scopes,
       },
       this.adapter.requester
     );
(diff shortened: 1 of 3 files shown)


packages/client/src/client.ts
```diff
@@ -443,7 +443,7 @@ export class StandardLogtoClient {
     }
 
     const accessTokenKey = buildAccessTokenKey(resource, organizationId);
-    const { appId: clientId } = this.logtoConfig;
+    const { appId: clientId, scopes } = this.logtoConfig;
```
wangsijie: Hi, I thought the plan was to add a "scope" parameter to the function `getAccessTokenByRefreshToken`. Reading `scopes` from `logtoConfig` does not really make sense here, because the `scopes` were already sent in the auth request with exactly the same value.

maddijoyce: Hey @wangsijie,

The issue for me was that I handle the sign in, etc and then call `getAccessToken()`, which returns an opaque access token. I follow that up with a call to `getAccessToken(resourceUrl)` in order to get a JWT with aud set to the resourceUrl. However, at the moment, that accessToken has no scopes, because they're not passed from the client's options. When I did the initial setup, before diving into the code, I assumed they would be because I had set them when I initialised the client.

The alternative would be to add the scopes as a third argument to `getAccessToken`

wangsijie: Hi @maddijoyce, I just did a test, init Logto client with resource (`https://my-resource.com/api`) and scopes (`profile`, `test`), and then call `getAccessToken('https://my-resource.com/api')`, the JWT contains the scope `test` which is expected. Can you please try again?

wangsijie: I can confirm that your configuration is correct and the user should be getting this scope. The issue might be that the user doesn’t have `test_scope` access on this resource, could you double check your RBAC configuration?

PR #1113, opened 2026-05-27
fix: cache OIDC discovery for server clients

### LINKED ISSUE #1112: @logto/next/server-actions instantiates a new LogtoClient per call, defeating the SDK's discovery memoisation and causing per-request fetches of    /oidc/.well-known/openid-configuration
## Summary

Every call to the top-level server-action functions in `@logto/next/server-actions` (e.g. `signIn`, `handleSignIn`, `getLogtoContext`, `getAccessToken`) constructs a fresh `LogtoClient` instance. The SDK's discovery-doc memoisation lives inside that instance (`once(this.#getOidcConfig)`), so it only ever fires once *per instance*. The net effect is that **every authenticated request to a Next.js app using this SDK triggers a fresh GET to `/oidc/.well-known/openid-configuration`** — and the same for `/oidc/me` for any call that requests userinfo.

Under realistic production load this saturates Logto Cloud's per-tenant rate limit on those endpoints.

## Root cause (source citations)

1. **[`packages/next/server-actions/index.ts:36`](https://github.com/logto-io/js/blob/master/packages/next/server-actions/index.ts#L36)** — fresh client per call:

   ```ts
   export async function signIn(config, options, interactionMode) {
     const client = new LogtoClient(config); // ← NEW client per call
     const { url } = await client.handleSignIn(finalOptions);
     redirect(url);
   }
   ```

   Same pattern at lines 59, 75, 99 for other server-action exports.

2. **[`packages/client/src/client.ts:82`](https://github.com/logto-io/js/blob/master/packages/client/src/client.ts#L82)** — memoisation is bound to the instance:

   ```ts
   readonly getOidcConfig: () => Promise<OidcConfigResponse> = once(this.#getOidcConfig);
   ```

3. **[`packages/client/src/utils/once.ts`](https://github.com/logto-io/js/blob/master/packages/client/src/utils/once.ts)** — `once` is implemented as a per-instance closure. New client → new closure → fresh `called = false`. The memoisation is technically present but useless once you stop reusing the client.

4. **`packages/client/src/adapter/index.ts`** — `adapter.unstable_cache` exists as a cross-instance persistence hook, but `@logto/next` / `@logto/node` don't wire one up.

## Related: `.json()` on non-JSON error bodies

Distinct but worth fixing in the same area:
**[`packages/client/src/utils/requester.ts:17`](https://github.com/logto-io/js/blob/master/packages/client/src/utils/requester.ts#L17)** calls `response.json()` on non-2xx responses without checking `Content-Type`:

```ts
if (!response.ok) {
  const cloned = response.clone();
  const responseJson = await response.json(); // ← throws on plain-text 429 body
  // ...
}
```

When Logto Cloud's rate limit fires, the body is `"Too many requests"` (text/plain), so consumers receive `SyntaxError: Unexpected token 'T', "Too many r"... is not valid JSON` — propagating past the typed-error catch handlers (`LogtoClientError` / `LogtoRequestError`).

## Impact

| Metric                                              | Before workaround        | After                                  |
| --------------------------------------------------- | ------------------------ | -------------------------------------- |
| 503 rate-limit responses to end users (75s @ 95 RPS) | 5,203 / 6,597 (78.9%)    | 0                                      |
| Upstream discovery fetches                          | ~6,597 (1:1 with requests) | 4 (~1 per Lambda per TTL window)       |

Production-tier traffic causes user-visible sign-in failures on any deployment using the server-actions API.

## Suggested fix (in order of preference)

1. **Module-scoped discovery cache.** Hoist the `LogtoClient` (or just the discovery result) to module scope inside `@logto/next` so all `signIn` / `handleSignIn` / etc. calls share one discovery fetch per process lifetime. The OIDC discovery doc is essentially static — fetching it once per process is the spec-recommended pattern.

2. **Expose a `discoveryDocument: OidcConfig` config option.** Lets consumers pre-populate discovery and skip the fetch entirely. Especially useful for serverless deploys where module scope is per-instance.

3. **Wire `adapter.unstable_cache` from `@logto/next` / `@logto/node`.** It already exists in the adapter contract; just needs a default implementation (e.g. an in-memory `Map` at module scope).

4. **Separately:** in `requester.ts`, check `response.headers.get('content-type')` before `.json()`. Construct a typed `LogtoRequestError` (with `code: 'rate_limited'` or similar) from non-JSON error bodies so consumers don't have to catch `SyntaxError`.

## Workaround for affected consumers (until a fix lands)

Patch `globalThis.fetch` at process start with a surgical interceptor that caches discovery + userinfo. Happy to share our implementation / extract a reusable utility package if useful for other teams.

## Environment

- `@logto/next`: 4.2.10
- `@logto/client`: (transitive — version per the linked source)
- Next.js: 16.x (App Router)
- Runtime: AWS Lambda via OpenNext
- Logto deployment: Logto Cloud with custom domain

### Expected behavior

The OIDC discovery document (/oidc/.well-known/openid-configuration) is essentially static per provider — its contents change only when an OIDC operator rotates keys, adds endpoints, or otherwise reconfigures the deployment. The OIDC spec explicitly assumes consumers will cache the response for the duration indicated by the response's Cache-Control header.

A correctly-behaving SDK should:

1. Fetch the discovery document once per process lifetime (or with a reasonable in-memory TTL, e.g. matching the response's Cache-Control header — typically minutes to hours), regardless of how many times signIn() / handleSignIn() / getAccessToken() etc. are invoked.
2. Share that cached discovery result across all LogtoClient instances within the same process — either by hoisting the client itself, by hoisting just the discovery cache, or by wiring adapter.unstable_cache to a module-scoped store.
3. As a result, a Next.js app handling N authenticated requests per process should make ~1 discovery fetch per process, not N. For comparison, other widely-used OIDC client libraries (e.g. openid-client / node-oidc-client) follow this pattern: discovery is performed once at issuer initialisation and the result is reused across all subsequent operations on that issuer.

## Reproduction

In a Next.js App Router app using `@logto/next/server-actions`:

```ts
// app/sign-in/route.ts
import { signIn } from "@logto/next/server-actions";
import { getLogtoConfig } from "@/lib/logto";

export async function GET() {
  await signIn(getLogtoConfig()); // ← fires discovery fetch
}
```

Load-test the route. Each request results in one upstream fetch to the discovery endpoint. At ~95 RPS we measured **78.9% of requests (5,203 / 6,597 in 75 s)** hitting Logto's rate limit and returning `429 Too many requests` with a plain-text body.

### Environment

Logto Cloud

### Screenshots

_No response_

wangsijie:
Fixes #1112

## Summary

- Cache OIDC discovery metadata across server-side client instances using endpoint-scoped in-memory storage.
- Deduplicate concurrent cache misses for shared cache storage.
- Normalize non-JSON request failures into typed Logto request errors.
- Document the adapter cache dedupe flow.

## Testing

Unit tests

### DIFF
--- a/.changeset/cache-oidc-discovery.md
+++ b/.changeset/cache-oidc-discovery.md
@@ -1,7 +1,6 @@
 ---
 '@logto/client': patch
 '@logto/node': patch
-'@logto/next': patch
 ---
 
 cache OIDC discovery metadata across server-side client instances and normalize non-JSON request failures
--- a/.changeset/cache-oidc-discovery.md
+++ b/.changeset/cache-oidc-discovery.md
@@ -1,0 +1,7 @@
+---
+'@logto/client': patch
+'@logto/node': patch
+'@logto/next': patch
+---
+
+cache OIDC discovery metadata across server-side client instances and normalize non-JSON request failures
--- a/packages/client/src/adapter/index.test.ts
+++ b/packages/client/src/adapter/index.test.ts
@@ -66,6 +66,17 @@ describe('ClientAdapterInstance', () => {
     await expect(firstCall).rejects.toThrow('transient failure');
     await expect(secondCall).resolves.toEqual({ test: 'recovered' });
     expect(secondGetter).toHaveBeenCalledTimes(1);
+    expect(await adapters.unstable_cache?.getItem(CacheKey.OpenidConfig)).toBe(
+      JSON.stringify({ test: 'recovered' })
+    );
+
+    const thirdGetter = vi.fn().mockResolvedValue({ test: 'should not run' });
+    const thirdAdapterInstance = new ClientAdapterInstance(adapters);
+
+    expect(await thirdAdapterInstance.getWithCache(CacheKey.OpenidConfig, thirdGetter)).toEqual({
+      test: 'recovered',
+    });
+    expect(thirdGetter).not.toHaveBeenCalled();
   });
 
   it('should deduplicate concurrent cache misses for the same cache storage', async () => {
--- a/packages/client/src/adapter/index.test.ts
+++ b/packages/client/src/adapter/index.test.ts
@@ -47,4 +47,25 @@ describe('ClientAdapterInstance', () => {
     expect(spy).toHaveBeenCalledTimes(2);
     expect(getter).toHaveBeenCalledTimes(1);
   });
+
+  it('should deduplicate concurrent cache misses for the same cache storage', async () => {
+    const adapters = createAdapters(true);
+    const firstAdapterInstance = new ClientAdapterInstance(adapters);
+    const secondAdapterInstance = new ClientAdapterInstance(adapters);
+    const getter = vi.fn(async () => {
+      await new Promise((resolve) => {
+        setTimeout(resolve, 10);
+      });
+
+      return { test: 'test' };
+    });
+
+    await expect(
+      Promise.all([
+        firstAdapterInstance.getWithCache(CacheKey.OpenidConfig, getter),
+        secondAdapterInstance.getWithCache(CacheKey.OpenidConfig, getter),
+      ])
+    ).resolves.toEqual([{ test: 'test' }, { test: 'test' }]);
+    expect(getter).toHaveBeenCalledTimes(1);
+  });
 });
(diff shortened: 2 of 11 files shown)


charIeszhao (approved):
LGTM

PR #1165, opened 2026-08-28
feat(angular): rebuild SDK on browser client

charIeszhao:
## Summary

- Replace the `angular-auth-oidc-client` configuration helper with a first-party Angular provider and service built on `@logto/browser`.
- Expose Angular Signals for authentication, loading, and error state together with Logto token, claims, userinfo, organization, sign-in, sign-out, and token-clearing APIs.
- Add hydration-safe post-render initialization, an overrideable client DI token, focused unit tests, and a major changeset for the 2.0.0 release.
- Rewrite the package README for v2 and add a v1-to-v2 migration guide, including replacing the comma-separated multi-resource workaround with the native `resources` array.

## Testing

Unit tests

## Checklist

- [x] `.changeset`
- [x] unit tests
- [ ] integration tests

### DIFF
--- a/packages/angular/src/provider.test.ts
+++ b/packages/angular/src/provider.test.ts
@@ -1,0 +1,75 @@
+import {
+  APP_INITIALIZER,
+  createEnvironmentInjector,
+  platformCore,
+  type EnvironmentInjector,
+  type Provider,
+} from '@angular/core';
+import LogtoClient from '@logto/browser';
+
+import { LOGTO_CLIENT, provideLogto } from './provider.js';
+import { LogtoService } from './service.js';
+
+const config = {
+  endpoint: 'https://logto.example',
+  appId: 'app-id',
+};
+
+const platform = platformCore();
+const platformInjector = platform.injector as EnvironmentInjector;
+const createInjector = (additionalProviders: Provider[] = []) =>
+  createEnvironmentInjector(
+    [provideLogto(config, { unstable_enableCache: true }), ...additionalProviders],
+    platformInjector
+  );
+
+describe('provideLogto', () => {
+  afterEach(() => {
+    localStorage.clear();
+    sessionStorage.clear();
+  });
+
+  afterAll(() => {
+    platform.destroy();
+  });
+
+  it('provides singleton client and service instances with the configured cache', () => {
+    const injector = createInjector();
+    const client = injector.get(LOGTO_CLIENT);
+    const service = injector.get(LogtoService);
+
+    expect(client).toBeInstanceOf(LogtoClient);
+    expect(client.logtoConfig).toMatchObject(config);
+    expect(client.adapter.unstable_cache).toBeDefined();
+    expect(injector.get(LOGTO_CLIENT)).toBe(client);
+    expect(injector.get(LogtoService)).toBe(service);
+    injector.destroy();
+  });
+
+  it('uses an Angular DI override for the underlying client', async () => {
+    const isAuthenticated = vi.fn(async () => true);
+    const client = { isAuthenticated } as unknown as LogtoClient;
+    const injector = createInjector([{ provide: LOGTO_CLIENT, useValue: client }]);
+
+    const service = injector.get(LogtoService);
+    await service.initialize();
+
+    expect(injector.get(LOGTO_CLIENT)).toBe(client);
+    expect(isAuthenticated).toHaveBeenCalledTimes(1);
+    expect(service.isAuthenticated()).toBe(true);
+    injector.destroy();
+  });
+
+  it('registers non-blocking browser initialization with Angular', () => {
+    const isAuthenticated = vi.fn(async () => true);
+    const client = { isAuthenticated } as unknown as LogtoClient;
+    const injector = createInjector([{ provide: LOGTO_CLIENT, useValue: client }]);
+
+    const initializers = injector.get(APP_INITIALIZER);
+
+    expect(initializers).toHaveLength(1);
+    expect(isAuthenticated).not.toHaveBeenCalled();
+    expect(injector.get(LogtoService).isLoading()).toBe(true);
+    injector.destroy();
+  });
+});
--- a/packages/angular/src/provider.test.ts
+++ b/packages/angular/src/provider.test.ts
@@ -2,8 +2,12 @@ import {
   APP_INITIALIZER,
   createEnvironmentInjector,
   platformCore,
+  provideZonelessChangeDetection,
+  runInInjectionContext,
   type EnvironmentInjector,
   type Provider,
+  ɵAfterRenderManager,
+  ɵINJECTOR_SCOPE,
 } from '@angular/core';
 import LogtoClient from '@logto/browser';
 
@@ -19,7 +23,12 @@ const platform = platformCore();
 const platformInjector = platform.injector as EnvironmentInjector;
 const createInjector = (additionalProviders: Provider[] = []) =>
   createEnvironmentInjector(
-    [provideLogto(config, { unstable_enableCache: true }), ...additionalProviders],
+    [
+      { provide: ɵINJECTOR_SCOPE, useValue: 'root' },
+      provideZonelessChangeDetection(),
+      provideLogto(config, { unstable_enableCache: true }),
+      ...additionalProviders,
+    ],
     platformInjector
   );
 
@@ -60,16 +69,28 @@ describe('provideLogto', () => {
     injector.destroy();
   });
 
-  it('registers non-blocking browser initialization with Angular', () => {
+  it('initializes after the first browser render without blocking app initialization', async () => {
     const isAuthenticated = vi.fn(async () => true);
     const client = { isAuthenticated } as unknown as LogtoClient;
     const injector = createInjector([{ provide: LOGTO_CLIENT, useValue: client }]);
 
     const initializers = injector.get(APP_INITIALIZER);
+    const service = injector.get(LogtoService);
 
     expect(initializers).toHaveLength(1);
     expect(isAuthenticated).not.toHaveBeenCalled();
-    expect(injector.get(LogtoService).isLoading()).toBe(true);
+    expect(service.isLoading()).toBe(true);
+
+    await runInInjectionContext(injector, async () => initializers[0]?.());
+
+    expect(isAuthenticated).not.toHaveBeenCalled();
+    injector.get(ɵAfterRenderManager).execute();
+
+    await vi.waitFor(() => {
+      expect(isAuthenticated).toHaveBeenCalledTimes(1);
+      expect(service.isAuthenticated()).toBe(true);
+      expect(service.isLoading()).toBe(false);
+    });
     injector.destroy();
   });
 });
--- a/packages/angular/src/service.ts
+++ b/packages/angular/src/service.ts
@@ -22,18 +22,19 @@ export class LogtoService {
   /** Whether initialization or one or more SDK operations are pending. */
   readonly isLoading: Signal<boolean>;
 
-  /** The latest SDK operation error, if any. */
+  /** The latest SDK operation error, cleared when another operation starts or by {@link clearError}. */
   readonly error: Signal<Error | undefined>;
 
   private readonly authenticatedState = signal(false);
-  private readonly loadingCount = signal(1);
+  private readonly initializationStarted = signal(false);
+  private readonly loadingCount = signal(0);
   // eslint-disable-next-line unicorn/no-useless-undefined -- Angular signal requires an initial value
   private readonly errorState = signal<Error | undefined>(undefined);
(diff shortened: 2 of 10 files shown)


packages/angular/src/service.ts
```diff
@@ -0,0 +1,197 @@
...
+      } catch (error: unknown) {
+        this.errorState.set(toError(error));
+      } finally {
+        this.stopLoading();
+      }
+    })();
+
+    this.initializationPromise = initializationPromise;
+    return initializationPromise;
+  }
```
simeng-li, line 74: `initializationPromise` is assigned unconditionally, including when `client.isAuthenticated()` rejected. Every later call returns the cached, already-resolved promise, so a failed initialization can never be retried — which contradicts the "operation is idempotent" note in the doc comment.

Concretely: if `isAuthenticated()` throws (blocked or corrupt `localStorage`, Safari private mode, an ID token that fails to decode), `error()` is set and `isAuthenticated()` is pinned to `false` for the rest of the session. A consumer that watches `error()` and calls `initialize()` again to recover gets the resolved promise straight back, so an already-signed-in user keeps seeing the signed-out UI with no way out short of a reload.

Clearing `this.initializationPromise` in the `catch` branch would make failures retryable.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,197 @@
...
+    interactionMode?: SignInOptions['interactionMode'],
+    loginHint?: SignInOptions['loginHint']
+  ): Promise<void> {
+    return this.run(async () => {
+      if (typeof options === 'string' || options instanceof URL) {
+        return this.client.signIn(options, interactionMode, loginHint);
+      }
+
+      return this.client.signIn(options);
+    }, true);
```
simeng-li: `client.signIn()` calls `clearAllTokens()` internally unless `clearTokens: false` is passed (`packages/client/src/client.ts:357`), and `client.signOut()` does the same at line 411 — but neither path updates `authenticatedState`. The service does sync it in `handleSignInCallback()` (true) and `clearAllTokens()` (false), so the mirrored state is inconsistent with itself.

Result: after a sign-in or sign-out attempt, `isAuthenticated()` still reports `true` while every token is gone. If the redirect doesn't complete — navigation blocked, `beforeunload` cancel, bfcache restore on Back — the app renders the authenticated UI, and any `getAccessToken()` from it throws `not_authenticated`.

I know `@logto/react` deliberately skips this for `signOut` (see `packages/react/src/hooks/index.ts:142-145`), but that reasoning was about avoiding a state flash *before* navigation. It doesn't cover the case where navigation never happens, and it doesn't explain the divergence from `clearAllTokens()` a few lines below.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,197 @@
...
+  }
+
+  /** Revoke local credentials and start the Logto sign-out redirect flow. */
+  async signOut(postLogoutRedirectUri?: string): Promise<void> {
+    return this.run(async () => this.client.signOut(postLogoutRedirectUri), true);
+  }
```
simeng-li, line 119: `keepLoadingOnSuccess = true` means the `startLoading()` in `run()` is never balanced by a `stopLoading()` here, so `loadingCount` stays permanently elevated and `isLoading()` can never return to `false` again in this document.

`window.location.assign()` only *starts* a navigation — the promise resolves while the page is still alive. If the sign-out navigation is aborted, or the user hits Back and the browser restores the SPA from bfcache, the app is stuck rendering the loading state forever, and there's no API on the service to reset the counter. The README template (`@if (logto.isLoading())`) hits this directly.

`@logto/react` resets loading after `signOut` for this reason — `proxy(client.signOut.bind(client))` uses the default `resetLoadingState = true`, and only `signIn` opts out. Can we match that here?

Note `service.test.ts:249` asserts `isLoading()` is `true` after sign-out, so the current test encodes the stuck state as expected — that assertion needs to flip too.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  /** Exchange the authorization callback for tokens and mark the session as authenticated. */
+  async handleSignInCallback(callbackUri: string): Promise<void> {
+    return this.run(async () => {
+      await this.client.handleSignInCallback(callbackUri);
+      this.authenticatedState.set(true);
+    });
+  }
```
simeng-li, line 127: The service forwards most of the client surface but not `isSignInRedirected()` (`packages/client/src/client.ts:370`), which is the guard `@logto/react` uses before handling a callback — `useHandleSignInCallback` checks it and only then calls `handleSignInCallback`.

Without it, the callback flow the README documents throws on any non-callback visit: the user signs in, lands on `/callback`, then refreshes or revisits the page; the sign-in session is already consumed, so `client.handleSignInCallback` throws `sign_in_session.not_found`. `run()` publishes that to the `error` signal — the README's `.catch(() => undefined)` hides the rejection but not the signal.

This becomes visible in #1166: `callback.component.ts` renders `logto.error()` in a `role="alert"` banner, so an already-signed-in user refreshing the callback page is shown a sign-in failure.

Exposing `isSignInRedirected()`, or adding an Angular equivalent of `useHandleSignInCallback`, would close this.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,197 @@
...
+
+  private stopLoading() {
+    this.loadingCount.update((count) => Math.max(0, count - 1));
+  }
+
+  private async run<Result>(
+    operation: () => Promise<Result>,
+    keepLoadingOnSuccess = false
+  ): Promise<Result> {
+    this.clearError();
```
simeng-li, line 201: This clears the error unconditionally at the start of every operation, so a later-starting concurrent operation wipes the error a previously-failed one just published.

It matters most for `initialize()`, which deliberately never rejects — the `error` signal is its only failure channel. If init fails and any component calls `fetchUserInfo()` / `getIdTokenClaims()` in the same tick, `run()` clears the signal and the initialization failure never surfaces anywhere. The reverse ordering drops errors too: a failed `getAccessToken()` is erased by `initialize()`'s own `clearError()` on line 51.

Could we clear only on an explicit `clearError()` call, or only when no other operation is in flight?

packages/angular/src/service.ts
charIeszhao: Updated in a3eab3ba. Failed initialization attempts are no longer cached, retries re-enter the loading state, concurrent callers still share one attempt, and a successful result remains cached. I also added regression coverage for failure, retry, loading, and successful caching.

packages/angular/src/service.ts
charIeszhao: I kept authentication state stable intentionally and documented that in a3eab3ba. This matches the React and Vue redirect behavior and avoids rendering an intermediate signed-out state before navigation. The explicit clearAllTokens method still updates the signal because it is the non-navigation local-clear API. Sign-out loading is settled separately.

packages/angular/src/service.ts
charIeszhao: Updated in a3eab3ba. signOut now uses the default loading cleanup, while only signIn keeps loading active for its navigation. The regression assertion now expects loading to settle.

packages/angular/src/service.ts
charIeszhao: Updated across a3eab3ba and stacked PR #1166. LogtoService now exposes isSignInRedirected, and both the README and sample check it before handling the callback. I added service delegation and callback-revisit coverage.

packages/angular/src/service.ts
charIeszhao: I kept the current behavior and clarified the public error semantics in a3eab3ba: error represents the latest operation and is cleared when a newer operation starts or through clearError. This is intentional retry-state behavior rather than error history; active operations still publish their own failure when they reject.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,214 @@
...
+        return false;
+      } finally {
+        this.stopLoading();
+      }
+    })();
+
+    this.initializationPromise = initializationPromise;
+    const isInitialized = await initializationPromise;
+
+    if (!isInitialized && this.initializationPromise === initializationPromise) {
```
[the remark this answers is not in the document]
charIeszhao: Simplified in 429167ca by clearing the cached promise based only on isInitialized. The focused tests, type check, and lint all pass.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,214 @@
...
+    // eslint-disable-next-line @typescript-eslint/unified-signatures -- preserve Browser client overloads
+    loginHint?: SignInOptions['loginHint']
+  ): Promise<void>;
+  async signIn(
+    options: SignInOptions | string | URL,
+    interactionMode?: SignInOptions['interactionMode'],
+    loginHint?: SignInOptions['loginHint']
+  ): Promise<void> {
+    // Keep authentication state stable while the redirect starts to avoid an intermediate UI state.
+    return this.run(async () => {
```
simeng-li: nit: this comment is about authentication state, but it sits directly above `run(..., true)`, whose argument controls the *loading* state. The identical sentence on `signOut` (line 108) reads correctly because there's no flag there — here it invites the reader to think `true` is what keeps auth state stable.

Maybe split it: keep this line for the `authenticatedState` decision, and put something like `// Keep loading active while the redirect navigates away.` next to the `true`.

packages/angular/src/service.ts
```diff
@@ -0,0 +1,214 @@
...
+
+  /** Revoke local credentials and start the Logto sign-out redirect flow. */
+  async signOut(postLogoutRedirectUri?: string): Promise<void> {
+    // Keep authentication state stable while the redirect starts to avoid an intermediate UI state.
+    return this.run(async () => this.client.signOut(postLogoutRedirectUri));
+  }
+
+  /** Check whether the current URL is the redirect URI for an active sign-in session. */
+  async isSignInRedirected(url: string): Promise<boolean> {
+    return this.run(async () => this.client.isSignInRedirected(url));
```
simeng-li: nit: routing this through `run()` means a pure predicate mutates shared state — it clears the `error` signal and toggles `isLoading()`.

In the README and sample flow that's harmless, since `isSignInRedirected()` runs first. But a router guard calling it to decide where to send the user would wipe an error another operation just published. Given you're keeping the "error reflects the latest operation" semantics, calling `this.client.isSignInRedirected(url)` directly would avoid adding a new clear-trigger — it only reads the sign-in session from storage and returns `false` rather than throwing on a cold visit.

packages/angular/src/provider.test.ts
```diff
@@ -0,0 +1,75 @@
...
+    const client = { isAuthenticated } as unknown as LogtoClient;
+    const injector = createInjector([{ provide: LOGTO_CLIENT, useValue: client }]);
+
+    const initializers = injector.get(APP_INITIALIZER);
+
+    expect(initializers).toHaveLength(1);
+    expect(isAuthenticated).not.toHaveBeenCalled();
+    expect(injector.get(LogtoService).isLoading()).toBe(true);
+    injector.destroy();
+  });
```
simeng-li: nit: this asserts the initializer is *registered* but never runs it, so the `provideAppInitializer` → `afterNextRender` → `initialize()` chain has no coverage anywhere in the suite.

Deleting the `void service.initialize()` call in `provider.ts` leaves all three tests in this file green: `initializers` still has length 1, `isAuthenticated` is still uncalled, and `isLoading()` is still `true` — now trivially so, since `initializationStarted` starts `false`. Line 72 in particular can't distinguish "non-blocking" from "never wired up", which is what the test name claims to cover.

The service-level tests added in a3eab3ba cover `initialize()` itself well, but the bootstrap wiring is the part specific to this SDK and it's what the PR's hydration-safety claim rests on. Could we run the initializer and flush the render hooks — `TestBed` plus `ApplicationRef.tick()`, or invoking the `APP_INITIALIZER` fn and awaiting it — and assert `isAuthenticated` actually gets called?

simeng-li (approved):
overall LGTM

packages/angular/src/service.ts
charIeszhao: Updated in 6d22a4e7. The authentication-state explanation remains above the operation, and a separate loading comment now sits directly next to the true flag.

packages/angular/src/service.ts
charIeszhao: Updated in 6d22a4e7. isSignInRedirected now delegates directly to the browser client, and the regression test verifies it preserves an existing error and leaves loading unchanged.

packages/angular/src/provider.test.ts
charIeszhao: Updated in 6d22a4e7. The provider test now invokes the registered APP_INITIALIZER in its injection context, flushes the Angular after-render manager, and verifies initialize calls isAuthenticated and settles the service signals.

PR #1170, opened 2026-09-08
feat(react-router): add request middleware

gao-sun:
#1169 introduced the coordinated session runtime for the React Router SDK. This second stacked change wires that runtime into React Router Framework Mode through a typed, request-scoped middleware API.

This middleware layer:

- adds `createLogtoReactRouter()` with `logto.middleware` and `logto.context`
- separates authentication context reads from explicit token acquisition
- checkpoints refresh-capable operations under the session coordinator
- propagates the final session through `Set-Cookie`
- supports custom coordination for multi-instance deployments
- covers loaders, actions, redirects, errors, and concurrent refreshes

The remaining stack will:

- migrate sign-in, callback, and sign-out routes to the middleware API
- remove the legacy React Router APIs
- update documentation and examples
- add the major-version changeset

### DIFF
--- a/packages/react-router/src/middleware/create-middleware.ts
+++ b/packages/react-router/src/middleware/create-middleware.ts
@@ -1,0 +1,59 @@
+import type { LogtoConfig } from '@logto/node';
+import type { MiddlewareFunction, RouterContext, Session, SessionStorage } from 'react-router';
+
+import { getCookieHeaderFromRequest } from '../framework/get-cookie-header-from-request.js';
+import { makeLogtoClient } from '../infrastructure/logto/create-client.js';
+import { createStorage } from '../infrastructure/logto/create-storage.js';
+import type { SessionCoordinator } from '../infrastructure/session/index.js';
+import { SessionRuntime } from '../infrastructure/session/index.js';
+
+import type { LogtoRequestContext } from './request-context.js';
+import { createLogtoRequestContext } from './request-context.js';
+
+type CreateLogtoMiddlewareOptions = Readonly<{
+  config: LogtoConfig;
+  context: RouterContext<LogtoRequestContext>;
+  sessionStorage: SessionStorage;
+  sessionCoordinator: SessionCoordinator;
+}>;
+
+const appendSessionCookie = (response: Response, cookieHeader: string | undefined) => {
+  if (!cookieHeader) {
+    return response;
+  }
+
+  const headers = new Headers(response.headers);
+  headers.append('Set-Cookie', cookieHeader);
+
+  return new Response(response.body, {
+    status: response.status,
+    statusText: response.statusText,
+    headers,
+  });
+};
+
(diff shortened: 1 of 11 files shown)


charIeszhao (approved):

packages/react-router/src/middleware/request-context.ts
```diff
@@ -0,0 +1,64 @@
...
+
+export type CreateLogtoRequestClient = (session: Session) => LogtoRequestClient;
+
+export const createLogtoRequestContext = (
+  runtime: SessionRuntime,
+  createClient: CreateLogtoRequestClient
+): LogtoRequestContext => {
+  const getAccessToken = async ({ resource, organizationId }: GetAccessTokenOptions = {}) =>
+    runtime.checkpoint(async (session) =>
+      createClient(session).getAccessToken(resource, organizationId)
```
charIeszhao, line 38: [P2] Preserve completed refresh-token rotation when ID-token verification fails

`getAccessToken()` can store a rotated refresh token before ID-token verification throws (see `packages/client/src/client.ts`, where `setRefreshToken()` precedes `verifyIdToken()`). If verification fails, for example because JWKS is temporarily unavailable, this checkpoint discards those mutations and `finalize()` has nothing to persist. Storage is left with the old credential, so a later refresh fails once that credential is no longer accepted. I reproduced the loss with the real Node client and an injected verification failure. Could we preserve completed token rotation on this error path while still propagating the verification failure?

packages/react-router/src/middleware/create-middleware.ts
```diff
@@ -0,0 +1,59 @@
...
+  return async ({ request, context }, next) => {
+    const runtime = await SessionRuntime.create({
+      cookieHeader: getCookieHeaderFromRequest(request) ?? undefined,
+      sessionStorage,
+      sessionCoordinator,
+    });
+    context.set(logtoContext, createLogtoRequestContext(runtime, createClient));
+
+    const response = await next();
+    const cookieHeader = await runtime.finalize();
```
charIeszhao, line 55: [P2] Wait for started session operations before finalizing a streamed response

A Framework Mode loader can return `{ token: context.get(logto.context).getAccessToken() }` and stream the unresolved promise. In that case, `next()` completes before the checkpoint, so `finalize()` returns no cookie even though token rotation later succeeds. With cookie-backed storage, the browser retains the old refresh token. I reproduced this through React Router 7.18.2's real `createRequestHandler` single-fetch pipeline: the response contains the resolved token but no `Set-Cookie`, while `commitSession()` runs after finalization. Could finalization wait for already-started session operations before returning the response?

charIeszhao (approved):
The two existing comments are non-blocking follow-ups.

wangsijie (approved):

charIeszhao (approved):

charIeszhao (approved):

PR #1187, opened 2026-09-10
feat(platform): expose token clearing APIs

gao-sun:
Expose explicit token-clearing operations through the platform SDK surfaces that wrap `@logto/node`, so applications do not need to reach into the underlying client.

- coordinated `clearAccessToken()` and `clearAllTokens()` on the React Router request context
- Pages Router and Server Actions helpers with writable cookie persistence
- Edge helpers that return updated response headers for cookie persistence

### DIFF
--- a/packages/next/edge/index.ts
+++ b/packages/next/edge/index.ts
@@ -123,6 +123,26 @@ export default class LogtoClient extends BaseClient {
   };
 
   /**
+   * Clear cached access tokens and return headers containing the updated session cookie. Attach
+   * the returned headers to the response so the cookie is persisted.
+   */
+  clearAccessToken = async (request: Request): Promise<Headers> => {
+    const { nodeClient, headers } = await this.createNodeClientFromEdgeRequest(request);
+    await nodeClient.clearAccessToken();
+    return headers;
+  };
+
+  /**
+   * Clear every locally stored token and return headers containing the updated session cookie.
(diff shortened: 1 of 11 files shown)


packages/next/edge/index.ts
```diff
@@ -122,6 +120,26 @@ export default class LogtoClient extends BaseClient {
     return context;
   };
 
+  /**
+   * Clear cached access tokens and return headers containing the updated session cookie. Attach
+   * the returned headers to the response so the cookie is persisted.
+   */
+  clearAccessToken = async (request: Request): Promise<Headers> => {
```
wangsijie: Both of these return bare `Headers`, while every other handler in this class returns a `Response`. If a caller drops the return value (easy to do, since TS won't complain), no `Set-Cookie` is sent and the tokens stay in the cookie. In middleware the headers also can't be passed straight to `NextResponse.next()` and have to be copied over by hand. Returning a `Response`, or taking the response/headers as an argument, removes that footgun. If the current shape is deliberate, a JSDoc usage example would help, since the README only links to the docs.

wangsijie (approved):

packages/next/edge/index.ts
gao-sun: Updated both methods to require writable response headers and mutate them directly, so callers no longer need to handle a returned `Headers` object. Added coverage in b5b9fb41.

Issue #1188, opened 2026-09-16
feature request: client-side access token refresh for SSR SDKs, starting with Nuxt


charIeszhao:
## Problem

SSR applications may need to call external APIs directly from the browser. An access token obtained during SSR and hydrated into client state eventually expires, but `@logto/nuxt` currently provides no built-in client-side API to obtain a fresh token without reloading the page or implementing a custom server endpoint.

Related: #970. That issue was closed after the reporter moved API calls to the Nuxt server; it did not add native client-side token refresh. Proxying API calls through the application server is useful, but does not cover applications that need browser-to-API requests.

## Proposed scope

Add SDK-supported client-side access token retrieval with on-demand refresh for SSR SDKs, starting with **Nuxt only**. Browser code should be able to request a valid access token throughout client-side navigation without a full-page reload or a hand-written token endpoint.

“Client-side refresh” means the browser can initiate token retrieval; the refresh-token exchange continues to run on the server using the existing authenticated session. Refresh tokens and application secrets must not be exposed to browser JavaScript.

The exact public API is open for design. A Nuxt composable backed by an SDK-provided server handler is one possible implementation.

## Initial investigation and reuse opportunities

Checked against master at `1cfbfd264316bb5f3d736a32db9737e0da7ba2e1`:

- **Nuxt:** [`useLogtoClient`](https://github.com/logto-io/js/blob/1cfbfd264316bb5f3d736a32db9737e0da7ba2e1/packages/nuxt/src/runtime/composables/use-logto-client.ts) reads SSR context and is unavailable in the browser. The [playground token endpoint](https://github.com/logto-io/js/blob/1cfbfd264316bb5f3d736a32db9737e0da7ba2e1/packages/nuxt/playground/server/api/auth/get-access-token.get.ts) demonstrates server-side retrieval, but is not registered by the SDK in consumer applications.
- **Shared foundation:** `@logto/node` inherits the [cached-token / refresh-token exchange logic in `@logto/client`](https://github.com/logto-io/js/blob/1cfbfd264316bb5f3d736a32db9737e0da7ba2e1/packages/client/src/client.ts). Reuse this instead of implementing a second refresh algorithm.
- **Next.js:** Already exposes token getters for server actions / API routes, and its Pages Router `handleUser` can return a context containing access tokens when configured. Its RSC getters explicitly cannot persist refreshed cookies. Future integration should build on these existing surfaces.
[...]

PR #1189, opened 2026-09-16
feat(nuxt): client-side access token retrieval with on-demand refresh

wangsijie:
## Summary

Adds `useLogtoAccessToken()` and an SDK-managed endpoint, so browser code can obtain a valid access token during client-side navigation without a page reload or a hand-written token endpoint.

Refs #1188.

## Design

The refresh still runs **on the server**. This is not a browser-held refresh token, and it does not expose the app secret:

- The endpoint is answered by the same request-scoped client the rest of the SDK uses, so an expired token is exchanged through the session and the renewed tokens are written to the response cookies before the response completes.
- Only the access token travels to the browser.
- The endpoint pathname is `pathnames.accessToken` (default `/api/logto/access-token`). The module mirrors it into `runtimeConfig.public.logto.accessTokenPath` so the composable and the handler cannot disagree about where the endpoint lives.
- `logtoEventHandler` answers the token request before resolving user info, so a token request never hits the userinfo endpoint.

The client construction moved to `runtime/utils/client.ts` (`createLogtoClient`) so the handler and the endpoint share it instead of duplicating it, and the response/error semantics live in `runtime/utils/access-token.ts` so they are unit-testable.

## API

```ts
const { accessToken, pending, error, refresh } = useLogtoAccessToken({
  resource: 'https://api.example.com',
  organizationId: 'org_123',
});

const token = await refresh();

if (!token) {
  // `not_authenticated`: the session cannot be renewed, so retrying would loop.
  await navigateTo('/sign-in');
}
```

`resource` and `organizationId` match the existing server-side `getAccessToken` parameters.

## Acceptance criteria

| Criterion | Implementation |
| --- | --- |
| Documented client-side API + SDK-managed server integration, with resource and organization parameters | `useLogtoAccessToken` + the endpoint in `logtoEventHandler`; README section added |
| Reuse an unexpired token, refresh an expired/missing one through the server session, persist the updated session before responding | The underlying client reuses its cached token and otherwise refreshes; `CookieStorage` writes the renewed session to the response cookies |
| Work after hydration and during client-side navigation without a reload | `refresh()` is called on demand in the browser |
| Concurrent calls must not cause duplicate refreshes | In-flight requests are shared per Nuxt app instance keyed by resource + organization; the state is shared through `useState` |
| Clear authentication error when the session is missing or refresh cannot succeed | `401` with `data.code = 'not_authenticated'`, exported as `NotAuthenticatedErrorCode` |
| Bind responses to the session, prevent caching, keep same-origin protections | `Cache-Control: no-store`; the existing httpOnly + `SameSite=Lax` session cookie authenticates the request, so no new cross-site surface |
| Update the guide and playground, cover expiry / persistence / concurrency / auth failures in tests | README section, reworked playground, 21 tests |

## Tests

```
test/access-token.test.ts             6 tests  response and error semantics
test/access-token-endpoint.test.ts    3 tests  routing + session persistence to Set-Cookie
test/use-logto-access-token.test.ts   9 tests  reuse, concurrency, error mapping, retry
test/event-handler.test.ts            3 tests  existing behaviour unchanged
```

Also verified against a running dev server: the endpoint answers `401` with `not_authenticated` when unauthenticated, and the playground page renders.

## Notes for review

1. **`refresh()` is browser-only.** During SSR the state stays empty, so no access token is ever serialized into the server-rendered payload. Server-side code uses `useLogtoClient()` as before. Making SSR work would mean putting the token in the HTML for hydration — I do not think that is worth it, but it is a design decision worth confirming.
2. **`invalid_grant` clears the session** before answering `401`, so the next request reports an unauthenticated user instead of replaying a doomed refresh. Other errors (for example `rate_limited`) are rethrown as server errors and clear nothing.
3. **Pathname resolution.** Nitro routes are build-time, so `pathnames.accessToken` is read when the module is set up. A runtime override therefore goes through `NUXT_PUBLIC_LOGTO_ACCESS_TOKEN_PATH`, which the handler and the composable both read. The JSDoc says so.
4. **Removed the playground's hand-written token endpoint** (`server/api/auth/get-access-token.get.ts`) since the SDK now provides it, and added `server/api/protected-resource.get.ts` as a stand-in resource server so the sample exercises a real browser-to-API call.


packages/nuxt/src/runtime/utils/handler.ts
```diff
@@ -88,10 +66,26 @@ export const logtoEventHandler = async (event: H3Event, config: RuntimeConfig) =
...
 
+  /**
+   * The access token request is answered before the user info is resolved: it only needs the
+   * session, and skipping the user info avoids an unnecessary request to the userinfo endpoint.
+   *
+   * The module mirrors the pathname into the public runtime config so that the composable and this
+   * handler cannot disagree about where the endpoint lives.
+   */
+  if (url.pathname === config.public.logto.accessTokenPath) {
+    return respondWithAccessToken(event, logto, getQuery<AccessTokenQuery>(event));
```
wangsijie: The payload returned here never reaches the browser. The handler Nitro registers, `runtime/server/event-handler.ts`, does `await logtoEventHandler(event, config)` and returns nothing, so this value is discarded. Route-less handlers are mounted as middleware (`nitropack/dist/runtime/internal/app.mjs`: `if (h.middleware || !h.route) h3App.use(...)`), and h3 1.15.11 only serializes a layer's return value when that layer returns it, so the request falls through to the next `/**` layer, which is Nuxt's renderer. `$fetch` gets the rendered 404 page instead of `{ accessToken }`, and `toAccessTokenError` surfaces `request_failed` even for a valid session. The 401 case still works because it throws, which is what the dev-server check exercised, and `test/access-token-endpoint.test.ts` calls `logtoEventHandler` directly so the route wrapper is never covered. Return the handler's result from `runtime/server/event-handler.ts`, or send the payload from here. Reproduced with h3 1.15.11 mounted the way Nitro mounts it: awaiting without returning falls through to the next layer.

packages/nuxt/src/runtime/utils/access-token.ts
```diff
@@ -0,0 +1,77 @@
...
+  event: H3Event,
+  client: LogtoClient,
+  { resource, organizationId }: { resource?: string; organizationId?: string } = {}
+): Promise<AccessTokenResponse> => {
+  if (!(await client.isAuthenticated())) {
+    throwNotAuthenticated();
+  }
+
+  try {
+    const accessToken = await client.getAccessToken(resource, organizationId);
```
wangsijie, line 111: Concurrent refreshes are deduped only per resource + organization within a single page, so two components asking for different resources (or the same token from two tabs) send parallel requests that each read the same refresh token from their request cookies. Refresh token rotation is on by default in Logto (for a confidential client like this one it kicks in once the refresh token passes 70% of its TTL), and reusing a consumed token trips reuse detection: the token endpoint answers `invalid_grant`, and the branch below then clears the whole session and responds 401, so a legitimate concurrent call signs the user out. `@logto/react-router` serializes token acquisition with `ProcessLocalSessionCoordinator` / `runtime.checkpoint` for exactly this reason.

wangsijie (commented):
Two findings inline. The endpoint's success payload is dropped by the server route wrapper, so only the 401 path can work; the unit tests call `logtoEventHandler` directly and miss it. Separate from that, refreshes are not serialized across concurrent requests, which can clear the session. CI is green.

packages/nuxt/src/runtime/utils/session-coordination.ts
```diff
@@ -0,0 +1,98 @@
...
+    } catch {
+      // The caller of `run` observes its own error; the queue only cares that it settled.
+    }
+  })();
+  sessionLocks.set(sessionKey, queued);
+
+  try {
+    return await run;
+  } finally {
+    if (sessionLocks.get(sessionKey) === queued) {
```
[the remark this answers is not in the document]
wangsijie: False positive: this is a deliberate identity comparison, not a value read. `queued` is compared with `===` to check whether this invocation is still the queue tail before removing the map entry; awaiting it would change the semantics and serve no purpose.

packages/nuxt/src/runtime/utils/session-coordination.ts
```diff
@@ -0,0 +1,98 @@
...
+
+/**
+ * Runs `operation` exclusively from other operations with the same session key. A rejected
+ * operation does not poison the queue for the next caller.
+ */
+export const runSessionExclusive = async <Result>(
+  sessionKey: string,
+  operation: () => Promise<Result>
+): Promise<Result> => {
+  const previous = sessionLocks.get(sessionKey) ?? Promise.resolve();
```
charIeszhao, line 99: [P2] Coordinate old and rotated cookies under one lock

After a refresh maps cookie A to B, a queued request with A still holds the A lock while a new request carrying B holds a separate B lock; both clients can therefore refresh from B concurrently. I reproduced both operations entering together with the current helpers. When rotation rejects the duplicate refresh with `invalid_grant`, `respondWithAccessToken` clears the session. Could we coordinate every cookie alias under one stable session identity and resolve the latest successor before refreshing?

packages/nuxt/src/runtime/utils/session-coordination.ts
```diff
@@ -0,0 +1,98 @@
...
+ * older cookie can adopt it. Every superseded value maps to the new one, including the value
+ * adopted from an earlier rotation, so chained refreshes cannot strand a request on a consumed
+ * refresh token.
+ */
+export const recordRotatedSessionValue = (supersededValues: string[], value: string): void => {
+  const expiresAt = Date.now() + rotatedSessionTtlMs;
+
+  for (const supersededValue of supersededValues) {
+    if (supersededValue !== value) {
+      rotatedSessions.set(supersededValue, { value, expiresAt });
```
charIeszhao: [P2] Evict expired rotation-cache entries independently of old-cookie lookups

The 30-second TTL does not remove entries unless the same superseded cookie is requested again. Once the browser adopts the new cookie, that lookup normally never happens, so every refresh leaves the old and new session strings retained for the lifetime of the server process. A clock-controlled reproduction retained all 1,000 expired mappings after a subsequent write and unrelated read. Could we bound this cache and sweep or schedule expiry independently of lookups?

charIeszhao (commented):
Two remaining session-coordination issues are detailed inline.

charIeszhao (approved):

PR #1190, opened 2026-09-16
feat(platform): support request-specific sign-in options

gao-sun:
Expose request-specific sign-in customization while keeping SDK-owned redirect fields authoritative.

- add `signInOptions` defaults and request-level `getSignInOptions` hooks to platform auth routes
- add a typed Nitro hook and event-handler resolver to Nuxt
- make Capacitor's object-form sign-in options work at runtime
- preserve sign-up registration routing and the legacy Next.js handler overload


charIeszhao (approved):
LGTM

packages/capacitor/src/index.ts
```diff
@@ -65,7 +65,12 @@ export default class CapacitorLogtoClient extends LogtoBaseClient {
     // system browser. We need to open an in-app browser to be able to handle
     // the redirects back to the app.
     // https://capacitorjs.com/docs/apis/browser
-    this.adapter.navigate = async (url) => {
+    this.adapter.navigate = async (url, parameters) => {
+      if (parameters.for === 'post-sign-in') {
+        window.location.assign(url);
+        return;
+      }
```
simeng-li, line 72: This makes `postRedirectUri` reachable in Capacitor for the first time, and it now navigates the WebView itself. That only works for an in-app route the WebView can load (e.g. `/home`). An external `https://` URL leaves the app shell or gets handed to the system browser, and a custom-scheme URL like `io.logto.example://home` fails to load on Android and is dropped on iOS. In both cases the user is authenticated but stranded with no error.

Please document this constraint on the object-form overload JSDoc and in the new README section (which currently says "the same object-form options as `@logto/browser`"), and note that it triggers a full WebView reload. The test example `https://app.example.com/after-sign-in` also models a value that would break on device.

packages/capacitor/src/index.ts
```diff
@@ -162,18 +170,21 @@ export default class CapacitorLogtoClient extends LogtoBaseClient {
...
         try {
-          // Kick off listener removal alongside the token exchange and browser dismiss
-          // (matches the original parallelism); the awaited Promise.all guarantees all
-          // three are done before resolve().
-          await Promise.all([this.handleSignInCallback(url), Browser.close(), cleanup()]);
+          // Browser dismissal is best-effort cleanup. It must not delay or override the token
+          // exchange, especially when the browser has already closed during deep-link handoff.
+          await Promise.all([
+            this.handleSignInCallback(url),
+            Browser.close().catch(swallowError),
```
simeng-li, line 191: `signOut` has the same shape at line 278 (`Promise.all([Browser.close(), cleanup()])`) and is still uncaught. The rationale in the comment above applies equally there: on the same deep-link handoff, `Browser.close()` rejects with `No active window to close!` and `signOut` rejects after tokens were already revoked and storage cleared. Could you mirror this `.catch(swallowError)` in `signOut` and add the corresponding test? There is currently no `signOut` test covering a rejecting `Browser.close()`.

.changeset/fresh-sign-in-options.md
```diff
@@ -0,0 +1,13 @@
...
+"@logto/nuxt": minor
+"@logto/next": minor
+"@logto/capacitor": minor
+---
+
+support request-specific sign-in options in platform SDKs
+
+Add static `signInOptions` defaults and per-request `getSignInOptions` callbacks to React Router,
+Express, Nuxt, and Next.js auth routes. Support the existing object-form sign-in options API in
+Capacitor, and use `firstScreen: 'register'` for the Next.js sign-up route.
```
simeng-li: Two other user-visible Capacitor changes in this PR are not mentioned here: `signIn` no longer rejects when `Browser.close()` fails (the `No active window to close!` case), and post-sign-in navigation now uses `window.location.assign` instead of opening the in-app browser. The first is a fix users are actively hitting, so it is worth a line so they can find the release that fixes it.

packages/capacitor/src/index.ts
```diff
@@ -130,12 +133,17 @@ export default class CapacitorLogtoClient extends LogtoBaseClient {
...
     interactionMode?: InteractionMode
   ): Promise<void> {
-    if (typeof redirectUri === 'object' && !(redirectUri instanceof URL)) {
-      throw new TypeError('The first argument must be a string or a URL.');
-    }
+    const options =
+      typeof optionsOrRedirectUri === 'string' || optionsOrRedirectUri instanceof URL
+        ? {
+            redirectUri: optionsOrRedirectUri,
+            ...(interactionMode && { interactionMode }),
```
simeng-li: nit: the conditional spread can be a plain `interactionMode,` key. The base client destructures it and `generateSignInUri` only appends the param when truthy, so `undefined` and absent are indistinguishable.

wangsijie (approved):

simeng-li (approved):

